PR #176, opened 2015-01-02
Use the Drupal 7 body_summary field as an excerpt if it's available

jessepinho:


lib/jekyll-import/importers/drupal7.rb
```diff
@@ -72,10 +74,11 @@ def self.process(options)
           # Get the relevant fields as a hash, delete empty fields and convert
           # to YAML for the header
           data = {
-             'layout' => 'post',
-             'title' => title.to_s,
-             'created' => created,
-           }.delete_if { |k,v| v.nil? || v == ''}.to_yaml
```
jessepinho: These lines were indented one extra space so I fixed it—hence the strange diff.

lib/jekyll-import/importers/drupal7.rb
```diff
@@ -72,10 +74,11 @@ def self.process(options)
...
+            'layout' => 'post',
+            'title' => title.to_s,
+            'created' => created,
+            'excerpt' => summary
```
parkr: Should `summary` be cast to a string? What are its expected values?

jessepinho: It can't hurt; but the `body_summary` field from the database is of type `LONGTEXT`, so I'm not sure it's necessary (especially considering `created` isn't being cast to an integer—it's just being trusted as a valid integer value).

jessepinho: If you'd feel more comfortable with a `.to_s`, though, I'm happy to add it.

PR #177, opened 2015-01-02
Remove unused nid

jessepinho:


parkr:
Think it's not useful? Works for me!

PR #189, opened 2015-03-07
Add post excerpt to WordPress.com importer

stephenyeargin:
Prior to this PR, the XML importer for WordPress.com would ignore the `post_excerpt` field. This is useful for SEO purposes (for the meta description in templates) as well as for other potential uses when displaying the content in a list view.


parkr:
LGTM!

PR #198, opened 2015-05-08
Update Usage documentation to match code example

douglashill:
Not sure if this should be on the master branch or gh-pages, so there’s a second pull request coming.

PR #199, opened 2015-05-08
Update Usage documentation to match code example

douglashill:
Not sure if this should be on the gh-pages branch or master. See PR #198 which makes the same change on master.


parkr:
On the master. :)

Issue #185, opened 2015-01-31
Wordpress Multisite


jk3us:
In trying to import a single site that is part of a wordpress multisite instance, I had to make 1 small change in the source.  In my case, I wanted to import the site that had the database table prefix `3_` (the id of the site is 3), but since multisite installations share a common set of users, and they are stored in the unprefixed `users` table, I just had to remove the #{px} for the users table on [line 144 of `importers/wordpress.rb`](https://github.com/jekyll/jekyll-import/blob/master/lib/jekyll-import/importers/wordpress.rb#L144), and it seems to have worked well.  Having a configuration option that could handle this case would be pretty nice.

parkr:
`--multisite`?

jk3us:
Is that a command-line switch? I don't see that in any docs or in the code. How would that be used?

parkr:
@jk3us I'm proposing it :) It'd do what you describe – use the specified table prefix except for the `users` table.

jk3us:
Ah, sounds good.  You'd need a way to specify which site, I guess by numeric id.  I think you can have the installation prefix in addition to the site prefix, e.g. wp_3_posts.

parkr:
@jk3us Hm, ok. I've never used a multisite before. Is the table prefix shared between them? So you have `prefix_num_table` and a separate `users` table?

jk3us:
Looks like it, here is the code that returns the database table prefix for a site.  http://doc-wordpress.com/nav.html?wp-includes/wp-db.php.source.html#l727

dirkkelly:
I have opened a pull request that will allow you to import specific sites from 3.4.2

https://github.com/jekyll/jekyll-import/pull/203

PR #203, opened 2015-06-24
Wordpress - Add Site Prefix 

dirkkelly:
I came across https://github.com/jekyll/jekyll-import/issues/185 when trying to import a multi site installation.

To give some background, the way that this installation has it configured, the sites are each stored as wp_x_ prefixes.

screen shot 2015-06-24 at 12 07 44 am

In order to support this I have added the `site-prefix` option which is prepended before all calls to these specific tables.

Import site `wp_2_`

```
ruby -rubygems -e 'require "jekyll-import";
    "dbname"      => "blogname",
    "user"        => "username",
    "password"    => "password",
    "host"        => "blog.com",
    "site_prefix" => "2_"
  })'
```

### DIFF
--- a/lib/jekyll-import/importers/wordpress.rb
+++ b/lib/jekyll-import/importers/wordpress.rb
@@ -19,7 +19,7 @@ module JekyllImport
         c.option 'password', '--password PW', "Database user's password (default: "")"
         c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
         c.option 'table_prefix', '--table_prefix PREFIX', 'Table prefix name (default: "wp_")'
-        c.option 'site_prefix', '--site_prefix PREFIX', 'Site prefix name (default: "2_")'
+        c.option 'site_prefix', '--site_prefix PREFIX', 'Site prefix name (default: "")'
         c.option 'clean_entities', '--clean_entities', 'Whether to clean entities (default: true)'
         c.option 'comments', '--comments', 'Whether to import comments (default: true)'
         c.option 'categories', '--categories', 'Whether to import categories (default: true)'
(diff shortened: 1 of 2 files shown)


lib/jekyll-import/importers/wordpress.rb
```diff
@@ -191,7 +198,6 @@ def self.process_post(post, db, options, page_name_list)
             excerpt = content[0...more_index]
           end
           if options[:more_anchor]
-            more_link = "more"
```
dirkkelly: vim pointed out this variable wasn't being used.

dirkkelly:
I have successfully used this on `3.4.2`, sorry I don't have a more recent installation.

lib/jekyll-import/importers/wordpress.rb
```diff
@@ -191,7 +198,6 @@ def self.process_post(post, db, options, page_name_list)
...
             content.sub!(/<!-- *more *-->/,
                          "<a id=\"more\"></a>" +
                          "<a id=\"more-#{post[:id]}\"></a>")
```
parkr: I think they wanted it here ^^

parkr:
Thanks for the patch! Just one comment. :)

dirkkelly:
Thanks @parkr, I've returned that line

parkr:
Great! Thanks, @dirkkelly. One more thing – can you please add this option to the documentation page for this importer?

dirkkelly:
@parkr I think I already have then? Sorry if I'm misunderstanding the request.

https://github.com/jekyll/jekyll-import/pull/203/files#diff-6ffabbc9b04e59e6903b73835e6b068bR29

parkr:
@dirkkelly Ah! Sorry about that. I was convinced we setup our documentation differently and I read that incorrectly. Thanks.

PR #208, opened 2015-07-07
Wordpress File Extension

dirkkelly:
Adds support for changing the extension for Wordpress posts.

Use case being I want my files as html.

```
  "extension" => "html"
})'
```


parkr:
I'm into it. Why not default to `html` too? I think we extract HTML from WordPress, so that would make more sense anyway.

dirkkelly:
@parkr I was surprised the default was markdown when it is pulling html. 

I'll change the default.

dirkkelly:
@parkr ready to go

parkr:
Thanks!

PR #212, opened 2015-07-24
Fix #211, wordpressdotcom import assets URLs

### LINKED ISSUE #211: jekyll-import wordpressdotcom xml - asset folder
When I import an xml file from a wordpress site, the posts import with assets rather than /assets - images and other files do not get linked to correctly in html files. The directory asset is linked to relatively, instead of to the root directory, which is where it sits. 

See more detail in this forum post - https://talk.jekyllrb.com/t/importing-wordpress-xml-downloads-images-to-assets-but-urls-to-this-dont-work/986/7

Excerpt from html file: 

```Img_20120302_195632</a>```

@gynter

### PR DESCRIPTION
**Pull request by gynter**
Currently URLs created for downloaded images in `assets_folder` are not prepended with slash therefore images are being linked as relative URLs i.e `/2012/03/16/assets/myimage.jpg` instead of the correct absolute ones `/assets/myimage.jpg`.

The fix also adds support for `site.baseurl` which means the images will also be linked correctly when serving site from custom base URL.

More info in [forums](https://talk.jekyllrb.com/t/importing-wordpress-xml-downloads-images-to-assets-but-urls-to-this-dont-work/986).

### DIFF
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -30,7 +30,7 @@ module JekyllImport
         images.each do |i|
           uri = i["src"]
 
-          i["src"] = assets_folder + "/" + File.basename(uri)
+          i["src"] = "{{ site.baseurl }}/%s/%s" % [assets_folder, File.basename(uri)]
           dst = File.join(assets_folder, File.basename(uri))
           puts "  " + uri
           if File.exist?(dst)

### CODE REVIEW

File: lib/jekyll-import/importers/wordpressdotcom.rb [hunk 1]
```diff
@@ -30,7 +30,7 @@ def self.download_images(title, post_hpricot, assets_folder)
         images.each do |i|
           uri = i["src"]
 
-          i["src"] = assets_folder + "/" + File.basename(uri)
+          i["src"] = "{{ site.baseurl }}/%s/%s" % [assets_folder, File.basename(uri)]
```
parkr: What do you think about:

```ruby
i["src"] = "{{ \"/#{File.join(assets_folder, File.basename(uri))}\" | prepend: site.baseurl }}"
```

?

danbernier:
:+1: I saw this same problem.

I _also_ saw some kind of "redirect not permitted" errors; looks like WordPress asset URLs are sent through a 302. I haven't thoroughly checked it out yet though.

lib/jekyll-import/importers/wordpressdotcom.rb
gynter: Yes, that would make more sense and it's a platform indenpendent way. I'll fix the PR tomorrow evening (EEST).

PR #213, opened 2015-08-10
Correctly handle wordpressdotcom drafts

danbernier:
When importing an unpublished draft from wordpress.com, jekyll-import was raising an exception trying to parse its published date, which is set to `0000-00-00 00:00:00`. It was using this to prepend the publish-date to the filename in `./_posts`.

This pull request changes the WordpressDotCom import so that:
* it doesn't try to parse the zero dates
* it doesn't prepend the date to the filenames
* it stores drafts in `./_drafts` instead of `./_posts`

I also added a bunch of test coverage for the WordpressDotCom importer, mostly by pulling lots of logic into a separate class, `JekyllImport::Importers::WordpressDotCom::Item`, and constructing those with specific XML. The actual writing of files is still untested, but there's less going on there now.

### DIFF
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -71,16 +71,16 @@ module JekyllImport
           end
         ] rescue {}
 
-        (doc/:channel/:item).each do |item|
-          title = item.at(:title).inner_text.strip
-          permalink_title = item.at('wp:post_name').inner_text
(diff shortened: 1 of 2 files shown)


parkr:
One of the best PR's ever! Some room for optimizations but looks solid to me! Thanks :)

danbernier:
You're welcome! It's a common refactoring I use :) Glad you memoized the methods, I'm surprised I forgot to do that.

PR #215, opened 2015-08-17
Use `test-unit` gem if using a version of Ruby where it's been removed

danbernier:
The `Test::Unit` classes have been removed from Ruby 2.2. You can still use them via the [test-unit gem](https://rubygems.org/gems/test-unit).

This adds the test-unit gem to the Gemfile, if the current Ruby version is after 2.1, the last Ruby version test-unit shipped with.


parkr:
I inadvertently just fixed this in #227. Sorry about that! The travis failures here discouraged me.

Issue #216, opened 2015-08-21
bad URI on Tumblr URLs with Umlauts


stefanw:
Tumblr puts umlauts in URLs, e.g like this:
http://blog.fragdenstaat.de/post/127162614837/nrw-gefängnis-muss-ehemaligem-strafgefangenem

Importing URLs with umlauts results in this exception:

```
ruby/2.1.0/uri/common.rb:176:in `split': bad URI(is not URI?): http://blog.fragdenstaat.de/post/125835141112/baden-württemberg-informationsfreiheitsgesetz (URI::InvalidURIError)
	from ruby/2.1.0/uri/common.rb:211:in `parse'
	from ruby/2.1.0/uri/common.rb:747:in `parse'
	from jekyll-import-0.7.1/lib/jekyll-import/importers/tumblr.rb:208:in `block in rewrite_urls_and_redirects'
	from jekyll-import-0.7.1/lib/jekyll-import/importers/tumblr.rb:204:in `map'
	from jekyll-import-0.7.1/lib/jekyll-import/importers/tumblr.rb:204:in `rewrite_urls_and_redirects'
	from jekyll-import-0.7.1/lib/jekyll-import/importers/tumblr.rb:61:in `process'
	from jekyll-import-0.7.1/lib/jekyll-import/importer.rb:23:in `run'
	from -e:2:in `<main>'
```

[This SO thread](https://stackoverflow.com/questions/10519591/how-to-use-umlauts-with-rubys-uri-library) advises to encode the URL before parsing, so maybe that could work here as well.

parkr:
Sounds good to me! Would you submit a PR?

PR #217, opened 2015-08-21
Encode tumblr URIs before parsing

stefanw:
They may contain umlauts and that breaks parsing, see #216.


parkr:
:+1: 

can you write a test for this, mayhaps?

PR #218, opened 2015-09-05
Fix post-merge hiccups after #213

danbernier:
It looks like there were some hiccups with the cleanup after merging #213:
* Some local variables were removed. This replaces their reference sites with calls to the appropriate Item methods.
* It was trying to call `.empty?` on an Hpricot::Doc, which doesn't support that method.
  * This also stops including empty excerpts in a Post's metadata.


parkr:
Thanks!

PR #221, opened 2015-10-11
Decode URI encoded multibyte file names

suzukimilanpaak:
I came across error saying a URI encoded file name is too long for a converted file name.
```
 ruby -rubygems -e 'require "jekyll-import";
quote>     JekyllImport::Importers::Blogger.run({
quote>       "source"                => "/Users/szkmp/Downloads/deletable/blog-10-10-2015.xml",
quote>       "no-blogger-info"       => false, # not to leave blogger-URL info (id and old URL) in the front matter
quote>       "replace-internal-link" => false, # replace internal links using the post_url liquid tag.
quote>     })'
/Users/szkmp/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/jekyll-import-0.8.0/lib/jekyll-import/importers/blogger.rb:187:in `initialize': File name too long @ rb_sysopen - _drafts/2013-01-23-%E3%83%87%E3%83%99%E3%83%AD%E3%83%83%E3%83%91%E3%81%8C%E3%83%87%E3%82%B6%E3%82%A4%E3%83%B3%E3%82%92%E3%81%99%E3%82%8B%E3%81%A8%E3%81%8D%E3%81%AB%E4%BE%BF%E5%88%A9%E3%81%9D%E3%81%86%E3%81%AA%E3%82%A6%E3%82%A7%E3%83%96%E3%82%B5%E3%82%A4%E3%83%88.html (Errno::ENAMETOOLONG)
	from /Users/szkmp/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/jekyll-import-0.8.0/lib/jekyll-import/importers/blogger.rb:187:in `open'
	from /Users/szkmp/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/jekyll-import-0.8.0/lib/jekyll-import/importers/blogger.rb:187:in `tag_end'
	from /Users/szkmp/.rbenv/versions/2.1.3/lib/ruby/2.1.0/rexml/parsers/streamparser.rb:28:in `parse'
	from /Users/szkmp/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/jekyll-import-0.8.0/lib/jekyll-import/importers/blogger.rb:46:in `block in process'
	from /Users/szkmp/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/jekyll-import-0.8.0/lib/jekyll-import/importers/blogger.rb:44:in `open'
[...]

PR #222, opened 2015-10-11
feat: convert blogger posts to markdown 

suzukimilanpaak:
Enable to convert blogger posts into markdown format if 'markdown' option is specified.

e.g.

```
ruby -rubygems -e ' require "jekyll-import"
    JekyllImport::Importers::Blogger.run({
      "source"                =>
"/Users/szkmp/Downloads/deletable/blog-10-10-2015.xml",
      "no-blogger-info"       => false, # not to leave blogger-URL info
(id and old URL) in the front matter
      "replace-internal-link" => false, # replace internal links using
the post_url liquid tag.
      "markdown" => true,
    })'
```

If the option not specified, it converts posts into html format as usual.

This is an ad-hoc way as it once converts posts into html stream then to markdown. It may be better to have another stream listener for this purpose. But, I thought this function is quite useful for many people and decided to send PR anyhow

NOTE: It's based on https://github.com/jekyll/jekyll-import/pull/221

Please review

### DIFF
--- a/lib/jekyll-import/importers/blogger.rb
+++ b/lib/jekyll-import/importers/blogger.rb
@@ -5,6 +5,7 @@ module JekyllImport
         c.option 'source', '--source NAME', 'The XML file (blog-MM-DD-YYYY.xml) path to import'
         c.option 'no-blogger-info', '--no-blogger-info', 'not to leave blogger-URL info (id and old URL) in the front matter (default: false)'
         c.option 'replace-internal-link', '--replace-internal-link', 'replace internal links using the post_url liquid tag. (default: false)'
+        c.option 'markdown', '--markdown', 'convert into markdown format'
       end
 
       def self.validate(options)
@@ -24,6 +25,8 @@ module JekyllImport
           time
           fileutils
           safe_yaml
+          open-uri
+          reverse_markdown
         ])
       end
 
@@ -32,6 +35,7 @@ module JekyllImport
       # source::                a local file String (or IO object for internal use purpose)..
       # no-blogger-info::       a boolean if not leave blogger info (id and original URL).
       # replace-internal-link:: a boolean if replace internal link
+      # markdown::              a boolean if convert into markdown format
       #
       # Returns nothing.
       def self.process(options)
@@ -39,6 +43,7 @@ module JekyllImport
 
         listener = BloggerAtomStreamListener.new
 
+        listener.markdown = options['markdown']
         listener.leave_blogger_info = ! options.fetch('no-blogger-info', false),
 
         File.open(source, 'r') do |f|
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -1,2 +1,4 @@
 source "https://rubygems.org"
 gemspec
+
+gem 'reverse_markdown'
```
parkr: You don't seem to have a test file, so you don't need this.

lib/jekyll-import/importers/blogger.rb
```diff
@@ -1,3 +1,6 @@
+require 'open-uri'
+require 'reverse_markdown'
```
parkr: Both of these need to be in `self.require_deps` function inside the `Blogger` class.

lib/jekyll-import/importers/blogger.rb
```diff
@@ -5,6 +8,7 @@ def self.specify_options(c)
         c.option 'source', '--source NAME', 'The XML file (blog-MM-DD-YYYY.xml) path to import'
         c.option 'no-blogger-info', '--no-blogger-info', 'not to leave blogger-URL info (id and old URL) in the front matter (default: false)'
         c.option 'replace-internal-link', '--replace-internal-link', 'replace internal links using the post_url liquid tag. (default: false)'
+        c.option 'markdown', '--markdown', 'convert into markdown format if specified'
```
parkr: please remove `if specific` from the description

parkr:
Three comments above, then it is good.

suzukimilanpaak:
@parkr fixed some points you mentioned. Please review again

lib/jekyll-import/importers/blogger.rb
```diff
@@ -183,20 +189,26 @@ def tag_end(tag)
...
+                file_name = URI::decode("#{post_data[:filename]}.#{extension}")
+                File.open(File.join(target_dir, file_name), 'w') do |f|
                   f.flock(File::LOCK_EX)
-      
+
                   f << post_data[:header].to_yaml
                   f << "---\n\n"
-                  f << post_data[:body]
+                  if markdown
+                    f << ReverseMarkdown.convert(post_data[:body])
```
parkr: I see the problem with the tests. Please add the require above this line, like so:

```ruby
if markdown
  JekyllImport.require_with_fallback(%w[reverse_markdown])
  f << ReverseMarkdown.convert(post_data[:body])
else 
  f << post_data[:body]
end
```

If it's in the `require_deps`, it's required even if you don't want to use ReverseMarkdown.

lib/jekyll-import/importers/blogger.rb
```diff
@@ -24,6 +25,8 @@ def self.require_deps
           time
           fileutils
           safe_yaml
+          open-uri
+          reverse_markdown
```
parkr: Please remove per my other comment.

parkr:
Thanks for all your work on this!

DirtyF:
👋 @suzukimilanpaak do you care addressing the feedback and fixing remaining conflicts? It would be a nice addition to this plugin.

DirtyF:
Feel free to re-open if you're still interested 😮

PR #223, opened 2015-10-30
now follows 'safe' http->https redirs for images

Bill:
The wordpress importer was failing to follow http->https redirections for my images (so it was failing to download them). This was due to open-uri not following these redirs.

Added the `open_uri_redirections` gem and changed the `open(uri)` call to:

``` ruby
open(uri, allow_redirections: :safe)
```

Also changed error logging method to say "Error" instead of "Errorr".

### DIFF
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('jekyll', '>= 1.4')
   s.add_runtime_dependency('fastercsv')
   s.add_runtime_dependency('nokogiri')
+  s.add_runtime_dependency('open_uri_redirections')
 
   # development dependencies
   s.add_development_dependency('rake', "~> 10.1.0")
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -11,6 +11,7 @@ module JekyllImport
           hpricot
           time
           open-uri
+          open_uri_redirections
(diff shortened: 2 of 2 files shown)


Bill:
Looks like the Travis CI failure is on Ruby 1.9.3:

`Gem::InstallError: jekyll requires Ruby version >= 2.0.0.`

I did not add the dependency on `jekyll`, so this must've been failing before my PR.

Bill:
Ah Jekyll 3.0.0 (the latest) requires Ruby 2:

https://rubygems.org/gems/jekyll/versions/3.0.0

Jekyll 3.0.0 was released on October 27, 2014 (three days ago). The previous version, 2.5.3, released in December of 2014 did not require Ruby 2.

Is it time to turn off the Ruby 1.9.3 tests in Travis CI?

parkr:
Not sure I love the added dependency. Can we do this manually?

mattr-:
IMHO, this isn't worth reimplementing ourselves. Why would we do this ourselves versus just adding the dependency?

parkr:
> IMHO, this isn't worth reimplementing ourselves. Why would we do this ourselves versus just adding the dependency?

Well having a new runtime dependency and asking for the dependency just for this one converter is different, I guess. My thought is to not have a runtime dependency. I guess we can ask for it for this converter, but redirects are so easy to do with Net::HTTP? We do a lot of that in jekyll-gist.

mattr-:
We do use a lot of Net::HTTP in jekyll-gist, but I didn't see any redirect handling code there.

PR #224, opened 2015-10-31
Add support for importing from PostgreSQL-backed Movable Type sites.

markstos:
Also, document the undocumented Movable Type importing options.

### DIFF
--- a/lib/jekyll-import/importers/mt.rb
+++ b/lib/jekyll-import/importers/mt.rb
@@ -26,10 +26,17 @@ module JekyllImport
       end
 
       def self.specify_options(c)
+        c.option 'engine', "--engine ENGINE", "Database engine, (default: 'mysql', postgres also supported)"
         c.option 'dbname', '--dbname DB', 'Database name'
         c.option 'user', '--user USER', 'Database user name'
         c.option 'password', '--password PW', "Database user's password, (default: '')"
         c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
+        c.option 'port', '--port PORT', 'Custom database port connect to (optional)'
+        c.option 'blog_id', '--blog_id ID', 'Specify a single Movable Type blog ID to import (default: all blogs)'
+        c.option 'categories', '--categories', "If true, save post's categories in its YAML front matter. (default: true)"
+        c.option 'src_encoding', '--src_encoding ENCODING', "Encoding of strings from database. (default: UTF-8)"
+        c.option 'dest_encoding', '--dest_encoding ENCODING', "Encoding of output strings. (default: UTF-8)"
+        c.option 'comments','--comments', "If true, output comments in _comments directory (default: false)"
       end
 
       # By default this migrator will include posts for all your MovableType blogs.
@@ -58,6 +65,7 @@ module JekyllImport
       def self.process(options)
         options  = default_options.merge(options)
 
+        engine   = options.fetch('engine', 'mysql')
         dbname   = options.fetch('dbname')
         user     = options.fetch('user')
         pass     = options.fetch('password', "")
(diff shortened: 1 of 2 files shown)


lib/jekyll-import/importers/mt.rb
```diff
@@ -58,15 +65,25 @@ def self.specify_options(c)
...
 
         posts_name_by_id = { } if comments
 
-        db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host)
+        if engine == 'mysql'
+          db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host, :port => port )
+        elsif engine == 'postgres'
+          db = Sequel.postgres(dbname, :user => user, :password => pass, :host => host, :port => port )
+        else
+          abort("unsupported --engine value supplied. Only 'mysql' and 'postgres' are supported")
```
parkr: Might be good to print out the value we got in quotes so that they can map that to the command they gave.

lib/jekyll-import/importers/mt.rb
```diff
@@ -58,15 +65,25 @@ def self.specify_options(c)
...
+        port     = options.fetch('port')
         comments = options.fetch('comments')
 
         posts_name_by_id = { } if comments
 
-        db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host)
+        if engine == 'mysql'
+          db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host, :port => port )
+        elsif engine == 'postgres'
+          db = Sequel.postgres(dbname, :user => user, :password => pass, :host => host, :port => port )
```
parkr: the arguments to these two Sequel calls are the same. You could always do `Sequel.public_send(engine, dbname, opts)` if you wanted

parkr:
Thank you! Can you please also investigate the Travis failure? If it's not due to your PR, just let me know.

markstos:
I've updated the PR as suggested and done a force push. 

The Travis failure is unrelated. It is due to to trying to test against Ruby 1.9.3, while Jekyll requires Ruby >= 2.0, so it appears the Ruby 1.9.3 tests will continue to fail dependably.

lib/jekyll-import/importers/mt.rb
```diff
@@ -58,15 +65,24 @@ def self.specify_options(c)
       def self.process(options)
         options  = default_options.merge(options)
 
+        engine   = options.fetch('engine')
```
parkr: this will need a default to match Line 29

lib/jekyll-import/importers/mt.rb
```diff
@@ -58,15 +65,24 @@ def self.specify_options(c)
...
         dbname   = options.fetch('dbname')
         user     = options.fetch('user')
         pass     = options.fetch('password', "")
         host     = options.fetch('host', "localhost")
+        port     = options.fetch('port')
```
parkr: `.fetch()` requires that a `port` is specified. to give it a default, do `options.fetch('port', 5432)` or, if it doesn't matter, you can just do `options['port']`.

parkr:
just a note about defaults ^^

parkr:
also, if you could rebase on the latest `master`. thank you! :pray:

lib/jekyll-import/importers/mt.rb
markstos: Fixed. Thanks for the catch.

lib/jekyll-import/importers/mt.rb
markstos: I think the issue you are highlighting here is that it might not work to pass  `:port => port` if `port` is not defined. I've updated the code so that now we only pass the `:port` key as a DB option if it is defined.    This seemed preferable to hardcoding the default ports for both MySQL and PostgreSQL into the code.

markstos:
OK, I've made some changes, checked them by re-running the importer, rebased the branch against `origin/master`, amended my commit and force-pushed my updated work back to my Github branch.

lib/jekyll-import/importers/mt.rb
parkr: That would be a side effect of changing this, but I believe you will get a KeyError if you do not specify port in the command line when you run this because the port key will not exist. See the description here:

http://ruby-doc.org/core-2.2.0/Hash.html#method-i-fetch

markstos: Thanks for your patience, I'm new to Ruby. 

I've confirmed you were right about my use of `fetch`, and have updated the code to just use `options['port']` as you originally suggested, as that appears to be the best fit for this optional value.

parkr:
Thanks a lot, @markstos! It's been released in v0.9.0.

PR #230, opened 2015-12-08
Allow importing any Drupal 7 node type

borfast:
This adds a `--types` command line option to allow the user to specify which node types to import from Drupal 7. This way the user can also import pages or any other content type.

It defaults to the ones the importer already did (blog posts, articles and stories), so that users are not forced to specify the node types and thus they won't see a difference if they don't need this ability.


parkr:
LGTM. Thank you!

borfast:
Thanks for merging! :)
I'll try to add the same functionality to the Drupal 6 importer as well.

PR #231, opened 2015-12-10
Allow importing any Drupal 6 node type

borfast:
Exactly the same thing as #230 but for Drupal 6 :)

This adds a --types command line option to allow the user to specify which node types to import from Drupal 6. This way the user can also import pages or any other content type.

It defaults to the ones the importer already did (blog posts, articles and stories), so that users are not forced to specify the node types and thus they won't see a difference if they don't need this ability.


borfast:
Thanks :)

parkr:
Thank you!!

PR #234, opened 2015-12-22
MT: allow use of SQLite for installation.

parkr:
Fixes #223

To use this patch, put the following in your `Gemfile`:

```ruby
gem "jekyll-import", github: "jekyll/jekyll-import", branch: "mt-use-sequel"
```

Then run `bundle install` and `bundle exec irb -rjekyll-import`. Then run the command per the docs specifying the `dbname` option as the path to your SQLite database file.

### DIFF
--- a/lib/jekyll-import/importers/mt.rb
+++ b/lib/jekyll-import/importers/mt.rb
@@ -2,6 +2,8 @@ module JekyllImport
   module Importers
     class MT < Importer
 
+      SUPPORTED_ENGINES = %{mysql postgres sqlite}
+
       STATUS_DRAFT = 1
       STATUS_PUBLISHED = 2
       MORE_CONTENT_SEPARATOR = '<!--more-->'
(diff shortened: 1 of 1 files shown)


lib/jekyll-import/importers/mt.rb
```diff
@@ -65,27 +67,10 @@ def self.specify_options(c)
...
-        if options['port']
-          db_opts[:port] = options['port']
-        end
-
-        if engine != 'mysql' && engine != 'postgres'
-          abort("unsupported --engine value supplied '#{engine}'. Only 'mysql' and 'postgres' are supported")
-        end 
-
-        db = Sequel.public_send(engine, dbname, db_opts)
+        db = database_from_opts(opts)
```
sixten: Should be `options`.

lib/jekyll-import/importers/mt.rb
```diff
@@ -243,6 +228,34 @@ def self.suffix(entry_type)
           entry_type
         end
       end
+
+      def self.database_from_opts(opts)
+        engine   = options.fetch('engine', 'mysql')
```
sixten: Should be `opts`.

lib/jekyll-import/importers/mt.rb
```diff
@@ -243,6 +228,34 @@ def self.suffix(entry_type)
...
+        dbname   = options.fetch('dbname')
+        user     = options.fetch('user')
+        pass     = options.fetch('password', "")
+        host     = options.fetch('host', "localhost")
+
+        db_connect_opts = {
```
sixten: Probably not much point building this outside of the MySQL/PostgreSQL case. Or fetching the user/password/host options, really.

parkr:
@sixten Would you mind trying again with my patch?

parkr:
@jekyllbot: merge +minor

PR #235, opened 2015-12-24
Clean up the Drupal importers

borfast:
Big refactoring of the Drupal importers.

All the code that was repeated on both Drupal6 and Drupal7 importers was pulled up into a generic `DrupalAbstract` class which implements the "core" logic of the Drupal importers and delegates the version-specific code to sub-classes.

I would appreciate it if someone could test importing non-Drupal content, though, because I don't have another blog to test the non-Drupal importers with.

### DIFF
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -34,8 +34,8 @@ module JekyllImport
       end
 
       def self.get_data(post)
-        content = post[:body]
-        summary = post[:teaser]
+        content = post[:body].to_s
+        summary = post[:teaser].to_s
         tags = (post[:tags] || '').downcase.strip
 
         data = {
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -38,8 +38,6 @@ module JekyllImport
         summary = post[:teaser]
         tags = (post[:tags] || '').downcase.strip
 
-        # Get the relevant fields as a hash, delete empty fields and convert
-        # to YAML for the header
         data = {
           :excerpt => summary,
           :categories => tags.split('|')
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -41,8 +41,8 @@ EOS
         tags = (post[:tags] || '').downcase.strip
 
         data = {
-          :excerpt => summary,
-          :categories => tags.split('|')
+          'excerpt' => summary,
+          'categories' => tags.split('|')
         }
 
          return data, content
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -10,21 +10,23 @@ module JekyllImport
         types = types.join("' OR n.type = '")
         types = "n.type = '#{types}'"
 
-        query = "SELECT n.nid, \
-                        n.title, \
-                        nr.body, \
-                        nr.teaser, \
-                        n.created, \
-                        n.status, \
-                        n.type, \
-                        GROUP_CONCAT( td.name SEPARATOR '|' ) AS 'tags' \
-                  FROM #{prefix}node_revisions AS nr, \
-                       #{prefix}node AS n \
-                  LEFT OUTER JOIN #{prefix}term_node AS tn ON tn.nid = n.nid \
-                  LEFT OUTER JOIN #{prefix}term_data AS td ON tn.tid = td.tid \
-                  WHERE (#{types}) \
-                    AND n.vid = nr.vid \
-                  GROUP BY n.nid"
+        query = <<EOS
+                SELECT n.nid,
+                       n.title,
+                       nr.body,
+                       nr.teaser,
+                       n.created,
+                       n.status,
+                       n.type,
+                       GROUP_CONCAT( td.name SEPARATOR '|' ) AS 'tags'
+                FROM #{prefix}node_revisions AS nr,
+                     #{prefix}node AS n
+                     LEFT OUTER JOIN #{prefix}term_node AS tn ON tn.nid = n.nid
+                     LEFT OUTER JOIN #{prefix}term_data AS td ON tn.tid = td.tid
+                WHERE (#{types})
+                  AND n.vid = nr.vid
+                GROUP BY n.nid
+EOS
 
         return query
       end
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -1,139 +1,53 @@
+require 'jekyll-import/importers/drupal_common.rb'
+
 module JekyllImport
   module Importers
     class Drupal6 < Importer
-      # Reads a MySQL database via Sequel and creates a post file for each story
-      # and blog node.
-      QUERY = "SELECT n.nid, \
-                      n.title, \
-                      nr.body, \
-                      n.created, \
-                      n.status, \
-                      GROUP_CONCAT( td.name SEPARATOR '|' ) AS 'tags' \
-                 FROM node_revisions AS nr, \
-                      node AS n \
-                 LEFT OUTER JOIN term_node AS tn ON tn.nid = n.nid \
-                 LEFT OUTER JOIN term_data AS td ON tn.tid = td.tid \
-                WHERE (%types%) \
-                  AND n.vid = nr.vid \
-             GROUP BY n.nid"
+      include DrupalCommon
+      extend DrupalCommon::ClassMethods
 
-      def self.validate(options)
-        %w[dbname user].each do |option|
-          if options[option].nil?
-            abort "Missing mandatory option --#{option}."
-          end
-        end
-      end
-
-      def self.specify_options(c)
-        c.option 'dbname', '--dbname DB', 'Database name'
-        c.option 'user', '--user USER', 'Database user name'
-        c.option 'password', '--password PW', "Database user's password (default: '')"
-        c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
-        c.option 'prefix', '--prefix PREFIX', 'Table prefix name'
-        c.option 'types', '--types TYPE1[,TYPE2[,TYPE3...]]', Array, 'The Drupal content types to be imported.'
-      end
-
-      def self.require_deps
-        JekyllImport.require_with_fallback(%w[
-          rubygems
-          sequel
-          fileutils
-          safe_yaml
-          mysql
-        ])
-      end
-
-      def self.process(options)
-        dbname = options.fetch('dbname')
-        user   = options.fetch('user')
-        pass   = options.fetch('password', "")
-        host   = options.fetch('host', "localhost")
-        prefix = options.fetch('prefix', "")
-        types  = options.fetch('types', ['blog', 'story', 'article'])
-
-        db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host, :encoding => 'utf8')
-
-        if prefix != ''
-          QUERY[" node "] = " " + prefix + "node "
-          QUERY[" node_revisions "] = " " + prefix + "node_revisions "
-          QUERY[" term_node "] = " " + prefix + "term_node "
-          QUERY[" term_data "] = " " + prefix + "term_data "
-        end
-
+      def self.get_query(prefix, types)
         types = types.join("' OR n.type = '")
-        QUERY[" WHERE (%types%) "] = " WHERE (n.type = '#{types}') "
+        types = "n.type = '#{types}'"
 
-        FileUtils.mkdir_p "_posts"
-        FileUtils.mkdir_p "_drafts"
-        FileUtils.mkdir_p "_layouts"
+        query = "SELECT n.nid, \
+                        n.title, \
+                        nr.body, \
+                        nr.teaser, \
+                        n.created, \
+                        n.status, \
+                        n.type, \
+                        GROUP_CONCAT( td.name SEPARATOR '|' ) AS 'tags' \
+                  FROM #{prefix}node_revisions AS nr, \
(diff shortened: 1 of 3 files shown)


parkr:
Great start, @borfast! I would love to see this as a module instead of a superclass, or have the one class handle both options and allowing folks to specify the version in the options, maybe. What do you think about that? That gets around the sub-subclassing issue and keeps the inheritance tree very clean.

borfast:
Thanks @parkr!

I actually thought about fixing the sub-subclassing issue by using modules instead but it turned out to cause more issues and since it required less code changes, I ended up going down the other route.

I do agree it's not pretty, though.

Having everything in one single class may not be the best option because it will get messy soon - there are already two Drupal importers and another one may soon pop up for Drupal 8. Having all these things in a single class sounds a bit too cramped.

But having a single Drupal class deriving from Importer, allowing the version to be specified which in turn dictates which specific module will be used sounds like a good idea, a much cleaner approach.

If you agree, I can take a stab at making it work that way.

borfast:
Hey @parkr, I finally got round to making the change into a module instead of a superclass.

How does that look to you?

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
+require 'jekyll-import/importers/drupal_common.rb'
```
parkr: You can drop the `.rb` from this. 😄

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
...
-      def self.specify_options(c)
-        c.option 'dbname', '--dbname DB', 'Database name'
-        c.option 'user', '--user USER', 'Database user name'
-        c.option 'password', '--password PW', "Database user's password (default: '')"
-        c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
-        c.option 'prefix', '--prefix PREFIX', 'Table prefix name'
-        c.option 'types', '--types TYPE1[,TYPE2[,TYPE3...]]', Array, 'The Drupal content types to be imported.'
+      def self.get_query(prefix, types)
+        types = types.join("' OR n.type = '")
+        types = "n.type = '#{types}'"
```
parkr: Can you add some documentation which defines what `types` is? And use different variable names for these intermediate values instead of reusing `types`?

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
...
+      extend DrupalCommon::ClassMethods
 
-      def self.specify_options(c)
-        c.option 'dbname', '--dbname DB', 'Database name'
-        c.option 'user', '--user USER', 'Database user name'
-        c.option 'password', '--password PW', "Database user's password (default: '')"
-        c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
-        c.option 'prefix', '--prefix PREFIX', 'Table prefix name'
-        c.option 'types', '--types TYPE1[,TYPE2[,TYPE3...]]', Array, 'The Drupal content types to be imported.'
+      def self.get_query(prefix, types)
```
parkr: We try not to use `get_` methods in Ruby; we prefer to see something like `query` for `get_query` and `query=` for `set_query`. In this case, your building a query. I'd be happy with something like `build_query`. 😄

lib/jekyll-import/importers/drupal6.rb
parkr: Alternatively, construct this piece of the query in its own method that can be called on line 26 below directly.

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
...
-      def self.require_deps
-        JekyllImport.require_with_fallback(%w[
-          rubygems
-          sequel
-          fileutils
-          safe_yaml
-          mysql
-        ])
+      def self.get_aliases_query(prefix)
+        return "SELECT src AS source, dst AS alias FROM #{prefix}url_alias WHERE src = ?"
```
parkr: Please avoid explicit `return`s. Just the string will do.

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
...
 
-      def self.require_deps
-        JekyllImport.require_with_fallback(%w[
-          rubygems
-          sequel
-          fileutils
-          safe_yaml
-          mysql
-        ])
+      def self.get_aliases_query(prefix)
```
parkr: Same as above about the `get_` prefix. Please remove it and just call this `aliases_query`.

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
...
-          # to YAML for the header
-          data = {
-             'layout' => 'post',
-             'title' => title.to_s,
-             'created' => created,
-             'categories' => tags.split('|')
-           }.delete_if { |k,v| v.nil? || v == ''}.each_pair {
-              |k,v| ((v.is_a? String) ? v.force_encoding("UTF-8") : v)
-           }.to_yaml
+      def self.get_data(post)
```
parkr: Same here with the `get_` prefix. Please call this `data`, or something which describes more of its purpose, like `build_post`. I'd recommend also renaming the `post` variable to `post_result` or something that indicates that it is from SQL.

lib/jekyll-import/importers/drupal7.rb
```diff
@@ -1,111 +1,54 @@
+require 'jekyll-import/importers/drupal_common.rb'
```
parkr: Same here with dropping the `.rb` prefix.

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -1,139 +1,53 @@
...
 
-          # Write out the data and content to file
-          File.open("#{dir}/#{name}", "w") do |f|
-            f.puts data
-            f.puts "---"
-            f.puts content
-          end
+        data = {
+          'excerpt' => summary,
+          'categories' => tags.split('|')
```
parkr: Why are the tags being assigned to the Jekyll post's `categories` flag? Shouldn't they remain `tags`?

lib/jekyll-import/importers/drupal7.rb
```diff
@@ -1,111 +1,54 @@
...
 
-      def self.require_deps
-        JekyllImport.require_with_fallback(%w[
-          rubygems
-          sequel
-          fileutils
-          safe_yaml
-        ])
+      def self.get_aliases_query(prefix)
+        return "SELECT source, alias FROM #{prefix}url_alias WHERE source = ?"
```
parkr: My comments from the drupal6.rb file also apply to this file 😄

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+          c.option 'types', '--types TYPE1[,TYPE2[,TYPE3...]]', Array, 'The Drupal content types to be imported.'
+        end
+
+        def require_deps
+          JekyllImport.require_with_fallback(%w[
+          rubygems
+          sequel
+          fileutils
+          safe_yaml
+        ])
```
parkr: Would you please fix the indentation here? Two spaces for each indentation, and `rubygems` and the rest of the list should be indented from `JekyllImport`.

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+      # and creates a post file for each node it finds in the Drupal database.
+
+      module ClassMethods
+        def specify_options(c)
+          c.option 'dbname', '--dbname DB', 'Database name'
+          c.option 'user', '--user USER', 'Database user name'
+          c.option 'password', '--password PW', "Database user's password (default: '')"
+          c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
+          c.option 'prefix', '--prefix PREFIX', 'Table prefix name'
+          c.option 'types', '--types TYPE1[,TYPE2[,TYPE3...]]', Array, 'The Drupal content types to be imported.'
```
parkr: Could you add the list of defaults here?

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+      #
+      # The general idea is that this importer reads a MySQL database via Sequel
+      # and creates a post file for each node it finds in the Drupal database.
+
+      module ClassMethods
+        def specify_options(c)
+          c.option 'dbname', '--dbname DB', 'Database name'
+          c.option 'user', '--user USER', 'Database user name'
+          c.option 'password', '--password PW', "Database user's password (default: '')"
+          c.option 'host', '--host HOST', 'Database host name (default: "localhost")'
```
parkr: It might make sense to specify these defaults as constants in the class and refer to them here and in `process`.

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+          dbname = options.fetch('dbname')
+          user   = options.fetch('user')
+          pass   = options.fetch('password', '')
+          host   = options.fetch('host', 'localhost')
+          prefix = options.fetch('prefix', '')
+          types  = options.fetch('types', %w(blog story article))
+
+          db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host, :encoding => 'utf8')
+
+          query = self.get_query(prefix, types)
```
parkr: This `self.` shouldn't be necessary.

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+
+          src_dir = Jekyll.configuration({})['source']
+
+          dirs = {
+              :_posts => File.join(src_dir, '_posts').to_s,
+              :_drafts => File.join(src_dir, '_drafts').to_s,
+              :_layouts => File.join(src_dir, '_layouts').to_s
```
parkr: If you're checking the configuration, it would be fruitful to check the value of `layouts_dir` here. It should default to `<source>/_layouts` but can be modified by the user.

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+          }
+
+          dirs.each do |key, dir|
+            FileUtils.mkdir_p dir
+          end
+
+          # Create the refresh layout
+          # Change the refresh url if you customized your permalink config
+          File.open(File.join(dirs[:_layouts], 'refresh.html'), 'w') do |f|
+            f.puts <<EOF
```
parkr: Please change this to `<<-HTML`.

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="content-type" content="text/html; charset=utf-8" />
+<meta http-equiv="refresh" content="0;url={{ page.refresh_to_post_id }}.html" />
+</head>
+</html>
+EOF
```
parkr: Please change this from `EOF` to `HTML`. :)

lib/jekyll-import/importers/drupal_common.rb
```diff
@@ -0,0 +1,148 @@
...
+          end
+
+          db[query].each do |post|
+            # Get required fields
+            data, content = self.get_data(post)
```
parkr: Can the contents of this `each` block please be pulled out into its own method and split up into sub methods if necessary? It's _huge_!

parkr:
@borfast Excellent, thank you! This is a terrific start. Just a few comments above. 😄 

Are you able to write tests for these as well?

borfast:
Sure thing, @parkr, I'll take care of these things as soon as possible. You can tell I don't use Ruby much :)

parkr:
@borfast Your Ruby is great! Much better than what I was writing before I was using it daily! 😄

parkr:
@borfast Want to patch the above or do you want me to do it?

borfast:
@parkr, sorry for taking so long, life hasn't been kind to me lately and I fell behind on several things.

I had already made a few of the easiest changes, which I just pushed in one commit. I'm not sure I'll have time for the rest in the next week, so if you want to finish up the rest of the stuff, go ahead.

lib/jekyll-import/importers/drupal6.rb
parkr: @borfast Would you mind answering this question for me? May just be a lexicographical difference between Jekyll and Drupal. 😄

borfast: @parkr, Drupal doesn't really distinguish between tags and categories, they're all "terms" in a "vocabulary". You can create multiple vocabularies and call them whatever you want, so if you feel like it, you can create a "Categories" vocabulary and a "Tags" vocabulary, each with its own set of terms that you can assign to each post.

The problem is, there's really no way of telling which vocabulary (if any) the author wants to use as the source for categories or tags because there's no hard-coded distinction between those two concepts. They're just "words" you assign to each post and they're grouped together under one or more collection of words.

The consequence is that it makes more sense to just assign everything to either categories or posts, instead of trying to guess which is which.

I suppose an option could be added to the importer to allow the user to specify which vocabularies should be used for tags and categories.

parkr:
LGTM!

parkr:
@borfast I have some incoming changes if you don't mind taking a look. They'll be in a follow-up PR.

@jekyllbot: merge

borfast:
Just replied on the other PR. Thanks for merging this one, @parkr :)

I have a few more ideas for improvements on the Drupal importer. I'll give them a shot when I have a bit more time.

parkr:
> I have a few more ideas for improvements on the Drupal importer. I'll give them a shot when I have a bit more time.

I will gladly review! Now that this monster is merged, merges should be quicker too 😄

PR #237, opened 2016-01-06
Add author and nid field to import from Drupal 7.

geerlingguy:
With the current Drupal 7 import script, there is no author data (very helpful when there are multi-author blogs, which is often the case with Drupal sites) nor nid attached to the imported posts. The nid can be extremely helpful in cases where you use a customized template to show Disqus comments or need to print other legacy-powered data on particular posts.

### DIFF
--- a/lib/jekyll-import/importers/drupal7.rb
+++ b/lib/jekyll-import/importers/drupal7.rb
@@ -7,12 +7,16 @@ module JekyllImport
                       fdb.body_value, \
                       fdb.body_summary, \
                       n.created, \
-                      n.status \
+                      n.status, \
+                      n.nid, \
+                      u.name \
                FROM node AS n, \
-                    field_data_body AS fdb \
+                    field_data_body AS fdb, \
+                    users AS u \
                WHERE (%types%) \
                AND n.nid = fdb.entity_id \
-               AND n.vid = fdb.revision_id"
+               AND n.vid = fdb.revision_id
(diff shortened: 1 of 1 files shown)


lib/jekyll-import/importers/drupal7.rb
```diff
@@ -7,12 +7,16 @@ class Drupal7 < Importer
...
                       fdb.body_summary, \
                       n.created, \
-                      n.status \
+                      n.status, \
+                      n.nid, \
+                      u.name \
                FROM node AS n, \
-                    field_data_body AS fdb \
+                    field_data_body AS fdb, \
+                    users AS u\
```
parkr: please add a space between `u` and `\`

parkr:
Wonderful! One nit above, then I'll merge :+1:

geerlingguy:
@parkr - Sorry about that, it's fixed now!

parkr:
Can this be backported to Drupal 6?

Thanks for your contributions!! :tada:

geerlingguy:
@parkr - The structure of the users table and uid/name fields is exactly the same between D6/D7... however, the field_data_body table and it's relationship to the node table isn't the same, so these particular changes could be backported, but the exact commit couldn't be cherry-picked into the drupal6 file/backported without rejiggering a little.

Unfortunately, I don't have any Drupal 6 sites I can use to test it, otherwise I'd do that really quick and backport it myself :/

PR #243, opened 2016-03-29
Tumblr: Update range of JSON readlines to extract JSON from JS (Tumblr may have changed structure of JSON)

karenjho:
Tumblr might have changed the structure of its JSON recently. The jekyll Tumblr importer is currently stripping out the ending curly brace (results in an "unexpected token error" with JSON parser).


parkr:
Thanks for the PR, @karenjho! What do you think about searching the string for the index of the first `{`, then finding the last `}`? Maybe not ideal, but a thought?

parkr:
@jekyllbot: merge +bug

PR #249, opened 2016-04-28
Tumblr: look up beginning and end of JSON dynamically

parkr:
Follow-up to #243.


parkr:
@jekyllbot: merge +bug

PR #251, opened 2016-05-14
Tumblr: fix stripping of JSONP characters from feed

vyuh:
should not try to read all lines twice from same stream (`feed`).
looked up docs for ruby arrays and ranges.


parkr:
Thanks!!

@jekyllbot: merge +bug

Issue #252, opened 2016-05-14
parse error, jekyll-import gem from tumblr


mattkhoury:
Trying to import my Tumblr site to Jekyll using the jekyll-import gem. Obviously, I'm not having much success. Any have any idea how to go about troubleshooting this error? Oh, and I'm new to Ruby - and - Jekyll, so, learning as I go.

`(JSON::ParserError)
	from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/json/common.rb:155:in `parse'
	from /Library/Ruby/Gems/2.0.0/gems/jekyll-import-0.10.0/lib/jekyll-import/importers/tumblr.rb:45:in `process'
	from /Library/Ruby/Gems/2.0.0/gems/jekyll-import-0.10.0/lib/jekyll-import/importer.rb:23:in `run'
	from -e:2:in `<main>'`

parkr:
Hey! Looking through our changelog, #243 and #249 may have fixed this. The error you're seeing is from malformed JSON. I can release a new version on Monday with the updated code. In the meantime, you can modify the code locally to pull in the JSON from the API as we do in #249. One way to do this is to copy the contents of tumblr.rb from GitHub here to your local gem (path is pasted above). Another way is to wait until Monday :-)

mattkhoury:
Hi Parkr, I'll give it another shot today. Sorry for the delay responding to you.

hairyhenderson:
This just happened to me, so I tried applying the fix from #249 into `tumblr.rb`, but there's an additional bug there that causes this:

```
Fetching http://myblog.tumblr.com/api/read/json/?num=50&start=0
/Users/me/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/json/common.rb:155:in `initialize': no implicit conversion of nil into String (TypeError)
	from /Users/me/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/json/common.rb:155:in `new'
	from /Users/me/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/json/common.rb:155:in `parse'
	from /Users/me/.rvm/gems/ruby-2.2.1/gems/jekyll-import-0.10.0/lib/jekyll-import/importers/tumblr.rb:48:in `process'
	from /Users/me/.rvm/gems/ruby-2.2.1/gems/jekyll-import-0.10.0/lib/jekyll-import/importer.rb:23:in `run'
[...]

PR #253, opened 2016-05-20
Tumblr: Fixing double-read and off-by-one error

hairyhenderson:
Found these bugs after I ran into #252 

Fixing two bugs introduced recently:
- `feed` was getting read from twice, so the second time it was empty
- `ending` marks the index of the right-most `}`, but needed to be
  bumped by 1 when referenced in the array

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/lib/jekyll-import/importers/tumblr.rb
+++ b/lib/jekyll-import/importers/tumblr.rb
@@ -43,8 +43,8 @@ module JekyllImport
           feed = open(feed_url)
           contents = feed.readlines.join("\n")
           beginning = contents.index("{")
-          ending = contents.rindex("}")
-          json = contents[beginning..ending]  # Strip Tumblr's JSONP chars.
+          ending = contents.rindex("}")+1
+          json = contents[beginning...ending]  # Strip Tumblr's JSONP chars.
           blog = JSON.parse(json)
           puts "Page: #{current_page + 1} - Posts: #{blog["posts"].size}"
           batch = blog["posts"].map { |post| post_to_hash(post, format) }


parkr:
LGTM, thank you. Ideally this would have a test – do you have interest in creating one?

hairyhenderson:
@parkr - thanks! Yeah, I have _interest_ in adding a test for this, but I'm not much of a Rubyist, so I'll have to bone up on the tests first probably :wink:...

parkr:
@hairyhenderson No problem! It'd probably come down to updating the [test `@jsonPayload`](https://github.com/jekyll/jekyll-import/blob/master/test/test_tumblr_importer.rb#L9) to the new JSON output from that call (what you see there is old) and passing that input to this method so it can be parsed.

You might want to create a method `self.extract_json` which receives `content` and returns the value of `blog`:

```ruby
# Parses the JSON from the JavaScript in the Tumblr API response
# and returns a Hash of its parsed data.
def self.extract_json(content)
  # ...
end
```

Then add a test which calls `extract_json` with the updated `@jsonPayload` variable and ensures you get the right thing in response. 😄

parkr:
I'll add the test in a minute. 👍 LGTM.

parkr:
@jekyllbot: merge

parkr:
@hairyhenderson Added a test for this in 28f0cad.

hairyhenderson:
Thanks @parkr - sorry for the lack of response! Was going to give it a shot, but _stuff_ got in the way :wink:

PR #258, opened 2016-05-28
Add functionality for importing Blogger comments

joscarsson:
Similar to the [WordPress importer](https://import.jekyllrb.com/docs/wordpress/), this PR adds support for importing comments from Blogger.

When `comments: true` is specified in `options` comments are written to a custom collection `_comments`. They are named according to the slug of the post, such as `2016-01-my-post-<number>.html` where `<number>` is a simple incrementing counter. To be able to connect comments with posts, there is a metadata field on the front matter `blogger_post_id` which refers to the post the comment is for (matches the `blogger_id` on posts).

Usage example can be seen on the blog I just imported, http://joscarsson.github.io/runtjordennu (warning, Swedish). The comments are added to the posts [through a generator](https://github.com/joscarsson/runtjordennu/blob/master/_plugins/static_comments.rb), and used [when rendering the posts](https://github.com/joscarsson/runtjordennu/blob/master/_layouts/post.html#L16-L27).

I don't usually code in Ruby, so this might be very non-standard/wrong - feel free to comment/give corrections.

### DIFF
--- a/lib/jekyll-import/importers/blogger.rb
+++ b/lib/jekyll-import/importers/blogger.rb
@@ -143,6 +143,10 @@ module JekyllImport
             if @in_entry_elem
               @in_entry_elem[:meta][:thumbnail] = attrs['url']
             end
+          when 'thr:in-reply-to'
+            if @in_entry_elem
+              @in_entry_elem[:meta][:post_id] = attrs['ref']
+            end
           end
         end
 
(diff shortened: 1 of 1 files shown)


parkr:
Neat! LGTM.

parkr:
@jekyllbot: merge

Issue #270, opened 2016-08-16
Documentation: Joomla3 is called Joomla


GreatEmerald:
In http://import.jekyllrb.com/docs/joomla3/ you can see the code specifying `JekyllImport::Importers::Joomla.run`, but that would call the Joomla 1 importer, not the Joomla 3 importer. That line should be changed into `JekyllImport::Importers::Joomla3.run` instead.
It looks like the file to be changed is https://github.com/jekyll/jekyll-import/blob/master/site/_importers/joomla3.md.

parkr:
Could you please submit a pull request to this effect/ Thanks!

GreatEmerald:
Sure, done.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

pathawks:
Fixed in #271 :+1:

PR #271, opened 2016-08-16
Fix for misnamed Joomla3 module in docs

GreatEmerald:
Fixes issue #270

### DIFF
--- a/site/_importers/joomla3.md
+++ b/site/_importers/joomla3.md
@@ -11,7 +11,7 @@ To import your posts from a [Joomla 3](http://joomla.org) installation, run:
 
 {% highlight bash %}
 $ ruby -rubygems -e 'require "jekyll-import";
-    JekyllImport::Importers::Joomla.run({
+    JekyllImport::Importers::Joomla3.run({
       "dbname"   => "name",
       "user"     => "myuser",
       "password" => "mypassword",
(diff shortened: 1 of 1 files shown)


pathawks:
Looks like we need to drop Ruby 1.9 from the test suite.

GreatEmerald:
Added another small improvement: the category field is numerical, not a string.

site/_importers/joomla3.md
```diff
@@ -11,18 +11,18 @@ To import your posts from a [Joomla 3](http://joomla.org) installation, run:
...
+      "category" => category,
       "prefix"   => "mytableprefix"
     })'
 {% endhighlight %}
 
 The only required fields are `dbname`, `prefix` and `user`. `password` defaults to `""`,
 and `host` defaults to `"localhost"`.
 
-If the `category` field is not filled, all articles will be imported, except the ones that are 
+If the `category` numerical field is not filled, all articles will be imported, except the ones that are
```
parkr: Is the category *always* numerical?

parkr:
Thanks! One question above.

GreatEmerald:
In the current implementation of the importer, yes. Joomla stores categories as numbers; there is also a table that maps those numbers into human-readable names. It would be nice to make use of that, but it would make the tool quite a bit more complex.

parkr:
@jekyllbot: merge +site

Issue #268, opened 2016-07-26
Joomla3: allow importing unpublished articles


GreatEmerald:
At the moment, the Joomla3 importer specifically filters out unpublished articles. Instead, it should import them into the `_drafts` directory, just like the wordpress importer does.

parkr:
@GreatEmerald That's a _great_ idea! Want to take a crack at it?

GreatEmerald:
I took a crack at it. It works, though it's quite basic.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

parkr:
This has a PR – just awaiting someone picking it up and addressing the comments.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

Dioda77:
_drafts

azzenabidi:
@parkr I would love to add this to my plate.

PR #272, opened 2016-08-17
joomla3: allow importing drafts

GreatEmerald:
Fix for issue #268. Also some whitespace cleanup (tabs and spaces where mixed beforehand).

This is a bit of a behaviour change in that it now only imports published articles by default, as opposed to both published and archived ones. But it can be rerun several times for several publishing states, so it shouldn't be much of an issue.

### DIFF
--- a/lib/jekyll-import/importers/joomla3.rb
+++ b/lib/jekyll-import/importers/joomla3.rb
@@ -16,6 +16,7 @@ module JekyllImport
         c.option 'host', '--host', 'Database host name'
         c.option 'category', '--category', 'ID of the category'
         c.option 'prefix', '--prefix', 'Table prefix name'
+        c.option 'state', '--state', 'Which publishing state to select. 0: unpublished, 1: published, 2: archived (default: 1)'
       end
 
       def self.require_deps
@@ -34,10 +35,16 @@ module JekyllImport
         host    = options.fetch('host', "localhost")
         cid	    = options.fetch('category', 0)
         table_prefix = options.fetch('prefix', "jos_")
+        state   = options.fetch('state', 1)
 
         db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host, :encoding => 'utf8')
 
-        FileUtils.mkdir_p("_posts")
+        if state == 0
+          directory = "_drafts"
+        else
+          directory = "_posts"
+        end
+        FileUtils.mkdir_p(directory)
 
         # Reads a MySQL database via Sequel and creates a post file for each
         # post in #__content that is published.
@@ -45,7 +52,7 @@ module JekyllImport
 		query << "`cn`.`created`, `cn`.`id`, `ct`.`title` AS `category`, `u`.`name` AS `author` "
 		query << "FROM `#{table_prefix}content` AS `cn` JOIN `#{table_prefix}categories` AS `ct` ON `cn`.`catid` = `ct`.`id` "
 		query << "JOIN `#{table_prefix}users` AS `u` ON `cn`.`created_by` = `u`.`id` "
-		query << "WHERE (`cn`.`state` = '1' OR `cn`.`state` = '2') " # Only published and archived content items to be imported
+		query << "WHERE `cn`.`state` = '#{state}' "
 
 		if cid > 0
 			query << " AND `cn`.`catid` = '#{cid}' "
@@ -58,12 +65,16 @@ module JekyllImport
           title = post[:title]
           slug = post[:alias]
           date = post[:created]
-		  author = post[:author]
-		  category = post[:category]
+          author = post[:author]
+          category = post[:category]
           content = post[:content]
-		  excerpt = post[:introtext]
-          name = "%02d-%02d-%02d-%s.markdown" % [date.year, date.month, date.day,
-                                                 slug]
+          excerpt = post[:introtext]
+          if state == 0
+            name = "%s.markdown" % [slug]
+          else
+            name = "%02d-%02d-%02d-%s.markdown" % [date.year, date.month, date.day,
+                                                   slug]
+          end
 
           # Get the relevant fields as a hash, delete empty fields and convert
           # to YAML for the header.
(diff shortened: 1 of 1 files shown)


lib/jekyll-import/importers/joomla3.rb
```diff
@@ -16,6 +16,7 @@ def self.specify_options(c)
         c.option 'host', '--host', 'Database host name'
         c.option 'category', '--category', 'ID of the category'
         c.option 'prefix', '--prefix', 'Table prefix name'
+        c.option 'state', '--state', 'Which publishing state to select. 0: unpublished, 1: published, 2: archived (default: 1)'
```
parkr: This will need to change to `state STATE` right after `c.option` so the parser knows it takes an arg. Surprised the ones above aren't like that... Whoops!

parkr: Also, I would consider using the full words here – it's _much_ more readable to use `--state unpublished` than `--state 0`.

lib/jekyll-import/importers/joomla3.rb
```diff
@@ -34,18 +35,24 @@ def self.process(options)
         host    = options.fetch('host', "localhost")
         cid	    = options.fetch('category', 0)
         table_prefix = options.fetch('prefix', "jos_")
+        state   = options.fetch('state', 1)
```
parkr: Here, you'd receive `state = "unpublished"`, which you could then map into the integer value using a hash:

```ruby
STATE_NUMS = {
  "unpublished" => 0, 
  # etc
}
```

lib/jekyll-import/importers/joomla3.rb
```diff
@@ -34,18 +35,24 @@ def self.process(options)
...
 
         db = Sequel.mysql(dbname, :user => user, :password => pass, :host => host, :encoding => 'utf8')
 
-        FileUtils.mkdir_p("_posts")
+        if state == 0
+          directory = "_drafts"
+        else
+          directory = "_posts"
+        end
```
parkr: Here you can squish this down to:

```ruby
directory = (state == "unpublished") ? "_drafts" : "_posts"
```

lib/jekyll-import/importers/joomla3.rb
```diff
@@ -34,18 +35,24 @@ def self.process(options)
...
+        FileUtils.mkdir_p(directory)
 
         # Reads a MySQL database via Sequel and creates a post file for each
         # post in #__content that is published.
         query = "SELECT `cn`.`title`, `cn`.`alias`, `cn`.`introtext`, CONCAT(`cn`.`introtext`,`cn`.`fulltext`) AS `content`, "
 		query << "`cn`.`created`, `cn`.`id`, `ct`.`title` AS `category`, `u`.`name` AS `author` "
 		query << "FROM `#{table_prefix}content` AS `cn` JOIN `#{table_prefix}categories` AS `ct` ON `cn`.`catid` = `ct`.`id` "
 		query << "JOIN `#{table_prefix}users` AS `u` ON `cn`.`created_by` = `u`.`id` "
-		query << "WHERE (`cn`.`state` = '1' OR `cn`.`state` = '2') " # Only published and archived content items to be imported
+		query << "WHERE `cn`.`state` = '#{state}' "
```
parkr: Here you'd use the hash we created above to grab the relevant integer value:

```ruby
query << "WHERE `cn`.`state` = '#{STATE_NUMS.fetch(state)}'"
```

lib/jekyll-import/importers/joomla3.rb
```diff
@@ -58,12 +65,16 @@ def self.process(options)
...
-		  excerpt = post[:introtext]
-          name = "%02d-%02d-%02d-%s.markdown" % [date.year, date.month, date.day,
-                                                 slug]
+          excerpt = post[:introtext]
+          if state == 0
+            name = "%s.markdown" % [slug]
+          else
+            name = "%02d-%02d-%02d-%s.markdown" % [date.year, date.month, date.day,
+                                                   slug]
+          end
```
parkr: Here I'd go with a string modification, actually:

```ruby
name = "#{slug}.markdown"
name.insert(0, "%02d-%02d-%02d" % [date.year, date.month, date.day]) unless state.eql?("unpublished")
```

parkr:
Thank you, @GreatEmerald! I took a look through and left a few comments. Let me know what you think! ✨

GreatEmerald:
Yeap, that sounds good generally. I was hoping someone with more Ruby knowledge would look at it ;)

I don't really get your first comment: what should all those options look like, then?

While I'm at it, I was also wondering if there's a good way to deal with https://github.com/jekyll/jekyll-import/blob/master/lib/jekyll-import/importers/joomla3.rb#L50 since it just filters out the "uncategorised" category, which may very well be useful to import; one can run the tool twice, once with category ID specifically set to 2, but, well, as you mentioned, that's pretty cryptic. Maybe an option like `--no-category` as the opposite for `--category` would be a better idea, but I don't really know how to make them mutually exclusive in a nice way... The same general method could be used for the drafts as well, allow excluding them instead of having to run the tool several times...

GreatEmerald:
Oh, and when I used the tool I noticed that some of the articles didn't actually get imported at all, despite me rerunning the tool four times to make sure nothing is exluded. I'll try to find out why.

parkr:
> Oh, and when I used the tool I noticed that some of the articles didn't actually get imported at all, despite me rerunning the tool four times to make sure nothing is exluded. I'll try to find out why.

Were you able to figure out this bug?

> Maybe an option like --no-category as the opposite for --category would be a better idea, but I don't really know how to make them mutually exclusive in a nice way... The same general method could be used for the drafts as well, allow excluding them instead of having to run the tool several times...

Let's think about the general use case for importing your posts into Jekyll. You probably want every piece of content you produces, including unpublished and archived. The only things I wouldn't care about are trashed items. When I run the importer, I'd like all of my draft, published, unpublished, and archived posts imported unless I specify. Perhaps the best thing is to default to importing *all* and only using the `--state` flag if I want a subset of that.

```console
$ jekyll import joomla3 
# Imports *everything* by default except trashed items
$ jekyll import joomla3 --states unpublished,published
# Imports posts with `unpublished` and `published` states only
```

What do you think?

GreatEmerald:
Yes, I double-checked and it turns out that it did import everything correctly after all, it was just me searching for one term when the article was named another...

And yes, that is indeed reasonable and how I expected the importer to work as well. The reason why I also made it initially only import one is because it makes code simpler, but the enum ("hash") idea would indeed nicely solve the issue to allow for importing all by default. As for the "uncategorised" category, indeed, I don't really know why the original author thought that it should be excluded by default. I'd rather it was included by default too.

DirtyF:
Closing after a long period of inactivity, feel free to re-open to feel conflicts and get this merged.

PR #280, opened 2016-10-18
Blogger: replace special characters with hyphens in the post name

### LINKED ISSUE #277: Blogger importer problem with special characters in title
I had several problems with filenames with the blogger importer. For example articles with `"` or `:` in its title.

After some investigation I found that the culprit was this part:

```
filename = "%s-%s" % [timestamp, CGI.escape(name.downcase).tr('+','-')]
```

I don't know ruby/ruby standard library so I did a dummy change to get it working based on this:
https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

```
filename = "%s-%s" % [timestamp, name.downcase.tr('+/\\:\'"<>{}?%*|.','---------------')]
```

Maybe ruby already provides something for that, I don't know. But CGI.escape doesn't seems to work fine for some names in this case.

soywiz:


soywiz:
Tried (with bash-for-windows and plain windows), but failed miserably, into setting up the test environment. So no tests here. I could try on my macbook but no guarantees. Also I don't know ruby.
Another option:
I could try to add a test that just places a xml with the title with special characters (using other test as reference), and do not place any assert at all. But at least checks that there are no exceptions. If I remember fine, it throws an exception with a wrong file name. And let travis tell me if that was fine directly in the PR :P

lib/jekyll-import/importers/blogger.rb
```diff
@@ -247,7 +247,7 @@ def get_post_data_from_in_entry_elem_info
               else
                 filename = "%s-%s" %
                   [timestamp,
-                   CGI.escape(name.downcase).tr('+','-')]
+                   name.downcase.tr('+/\\:\'"<>{}?%*|.','-')]
```
parkr: We might still need the CGI.escape here. I might also put this into its own method.

```ruby
def sanitized_filename(timestamp, name)
  return timestamp if name.nil?

  "%s-%s" % [
    timestamp,
    CGI.escape(name.downcase.tr('+/\\:\'"<>{}?%*|.', '-'))
  ]
end
```

Then remove from `if name.nil?` to the first `end`, and replace it with `filename = sanitized_filename(name, timestamp)` 😄

parkr:
Fixes #277.

parkr:
@soywiz Do you still want to apply the changes that I asked for above?

parkr:
Thank you! 

@jekyllbot: merge +bug

PR #283, opened 2016-11-29
Tumblr: fix creation of rewrite rules

### LINKED ISSUE #256: Tumblr import error
Hello, I'm running into problems that look very similar to what was reported in #252.  Like the OP for that issue, I am new to both Ruby and Jekyll.  I'm running Ruby 2.2 on Windows 10 (64-bit).  I selected Ruby 2.2 because I wasn't able to install jekyll-import without reverting to an earlier version of Ruby.

When I attempt to run the importer, I get screenfuls of HTML followed by this error:

`(JSON::ParserError)
    from c:/Ruby22/lib/ruby/2.2.0/json/common.rb:155:in`parse'
    from c:/Ruby22/lib/ruby/gems/2.2.0/gems/jekyll-import-0.10.0/lib/jekyll-import/importers/tumblr.rb:46:in `process'
    from c:/Ruby22/lib/ruby/gems/2.2.0/gems/jekyll-import-0.10.0/lib/jekyll-import/importer.rb:23:in`run'
    from import.rb:3:in `<main>'`

As this error appears to be very similar to the one reported in #252 I tried some of the proposed fixes, including downloading the most recent version of `tumblr.rb` from this repository.  But I wasn't successful and I'm actually skeptical that I implemented the proposed fixes, replacing some lines of code in `tumblr.rb`, correctly.

I'd appreciate any pointers to the fix or any advice about questions to ask.

Thank you.

parkr:
Fixes #256.

### DIFF
--- a/lib/jekyll-import/importers/tumblr.rb
+++ b/lib/jekyll-import/importers/tumblr.rb
@@ -213,12 +213,12 @@ module JekyllImport
         urls = Hash[posts.map { |post|
           # Create an initial empty file for the post so that
           # we can instantiate a post object.
+          File.write("_posts/tumblr/#{post[:name]}", "")
           tumblr_url = URI.parse(URI.encode(post[:slug])).path
           jekyll_url = if Jekyll.const_defined? :Post
-                         File.open("_posts/tumblr/#{post[:name]}", "w") { |f| f.puts }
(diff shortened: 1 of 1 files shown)


parkr:
@jekyllbot: merge +bug

PR #286, opened 2016-12-21
Corrected Joomla3 importer name and clarified "category" field

Pxlatn:
The current documentation implies that the "category" field requires a string when it actually needs the id of the category to be selected.


parkr:
Thank you!

@jekyllbot: merge +site

PR #287, opened 2017-01-15
Drupal 7: Remove uneeded double quote in SQL query

njt1982:
Using EOS around the query means we dont need the trailing quote.


parkr:
@jekyllbot: merge +bug

Issue #288, opened 2017-01-15
Sequel::DatabaseError: Mysql::Error: Expression #3 of SELECT list is not in GROUP BY clause


njt1982:
This might be an issue with default configs of newer MySQL/Percona setups..

```
Sequel::DatabaseError: Mysql::Error: Expression #3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'drupal7.fdb.body_value' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

njt1982:
See PR #289

PR #289, opened 2017-01-15
Fixes SQL grouping error in #288

njt1982:
See issue #288...

```
Sequel::DatabaseError: Mysql::Error: Expression #3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'drupal7.fdb.body_value' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```


njt1982:
This also removes the need for #287...

parkr:
Looks good to me!

@jekyllbot: merge +bug

PR #290, opened 2017-01-15
Updating Ruby and Jekyll versions for testing

njt1982:
Nokogiri no longer supports Ruby 1.9.3 it seems, causing tests to fail.

Given [1.9.3 is EOL](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/), IMHO it should not longer be supported by this Gem.

Also, [2.0 is EOL](https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/) so there is no point adding that. And it seems [2.3 and 2.4](https://docs.travis-ci.com/user/languages/ruby/#Supported-Ruby-Versions-and-RVM) are not available yet either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,8 @@ env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   matrix:
-    - JEKYLL_VERSION=3.3
     - JEKYLL_VERSION=3.0
-    - JEKYLL_VERSION=2.5.3
+    - JEKYLL_VERSION=2.5
 cache: bundler
 sudo: false
 before_script: bundle update
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 rvm:
+  - 2.3.1
   - 2.2
   - 2.1
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ env:
   matrix:
     - JEKYLL_VERSION=3.3
     - JEKYLL_VERSION=3.0
-    - JEKYLL_VERSION=2.5
-    - JEKYLL_VERSION=2.0
+    - JEKYLL_VERSION=2.5.3
 cache: bundler
 sudo: false
 before_script: bundle update
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 2.3
   - 2.2
   - 2.1
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,16 @@
 language: ruby
 rvm:
+  - 2.3
   - 2.2
   - 2.1
-  - 2.0
-matrix:
-  include:
-    - # Ruby 1.9
-      rvm: 1.9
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 2.3
```
pathawks: [Try `2.3.1`](https://github.com/jekyll/jekyll/blob/v3.3.1/.travis.yml#L8)

.travis.yml
```diff
@@ -2,18 +2,14 @@ language: ruby
...
-      rvm: 1.9
-      env: JEKYLL_VERSION=2.0
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   matrix:
+    - JEKYLL_VERSION=3.3
     - JEKYLL_VERSION=3.0
     - JEKYLL_VERSION=2.5
+    - JEKYLL_VERSION=2.0
```
pathawks: Is there a reason to support Jekyll 2.0?
You could probably even drop Jekyll 2.5 at this point.

.travis.yml
njt1982: I'll try 2.3.1 if a653148 passes :)

.travis.yml
njt1982: I was trying to find what Jekyll versions were being supported. I guess now 3.x is up to 3.3, the 2.x branch is probably dead? Should be tested for legacy users I suppose... 

Does 2.5 uses the [2.5.3 release](https://github.com/jekyll/jekyll/releases/tag/v2.5.3)?

.travis.yml
njt1982: @pathawks looks like 2.3.1 is supported - https://travis-ci.org/jekyll/jekyll-import/jobs/192211050

.travis.yml
```diff
@@ -1,19 +1,15 @@
...
-matrix:
-  include:
-    - # Ruby 1.9
-      rvm: 1.9
-      env: JEKYLL_VERSION=2.0
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   matrix:
+    - JEKYLL_VERSION=3.3
```
pathawks: I think Bundler actually resolves 3.0 to the latest 3.x version, as can be seen [here](https://travis-ci.org/jekyll/jekyll-import/jobs/192207963#L313)

.travis.yml
```diff
@@ -1,19 +1,15 @@
...
     - JEKYLL_VERSION=3.0
-    - JEKYLL_VERSION=2.5
+    - JEKYLL_VERSION=2.5.3
```
pathawks: Same as above, but if we want to be extra explicit, I don't think there's any harm :+1:

njt1982: Ahh cool - I think using the lastest within that branch makes more sense. Adds maintenance headaches otherwise.

njt1982:
@pathawks thanks for the quick reviews - would be good to get this in so other PR's can be properly tested :)

parkr:
@jekyllbot: merge +dev

PR #292, opened 2017-01-28
WordPress: replace was backwards, broke end tags

### LINKED ISSUE #254: WordPress import is improperly encoding HTML end tags
I just imported my Wordpress site (self-hosted) and all HTML end tags had the `"/"` replaced with `"&#47;"`, which of course makes them ever so slightly less useful as HTML end tags.

I commented the following line in `wordpress.rb` and it worked, but since I don't know the background of why this line was added in the first place (there must be a reason), I didn't want to send a naive pull request:

`text.gsub!("/", "&#47;")`

Using the latest Jeckyll, jeckyll-import, Wordpress, etc.

richardtallent:
The substitution was not being made because the arguments were reversed. This resulted in breaking HTML end tags imported from WordPress. The issue reporting this was closed due to inactivity, I've just now had some time to try to use Jekyll again.

### DIFF
--- a/lib/jekyll-import/importers/wordpress.rb
+++ b/lib/jekyll-import/importers/wordpress.rb
@@ -346,7 +346,7 @@ module JekyllImport
         text.gsub!("&gt;", ">")
         text.gsub!("&quot;", '"')
         text.gsub!("&apos;", "'")
-        text.gsub!("/", "&#47;")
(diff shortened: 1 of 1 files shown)


parkr:
Nicely done!

Fixes #254.

@jekyllbot: merge

PR #299, opened 2017-04-02
F/s9y db

till:
Tried to use the s9y database import and found two issues:

 - couldn't connect to a custom MySQL host (kind of exposed when you ssh-tunnel from local to remote)
 - it didn't import the "extended" portion of the blog entry


parkr:
@jekyllbot: merge +bug

till:
🎉  Thank you!

PR #304, opened 2017-04-16
Ghost: import drafts & pages, and fix issue with date extraction

proog:
This PR adds a few improvements to the Ghost importer.

 - The note about weird Ghost date formats seems to no longer apply. When run against a current Ghost installation (tested with 0.11), all dates end up as 1970-01-01 because of the special trimming. This has been removed.
 - If a post has been published, the importer now uses its publish date instead of its creation date.
 - The importer now imports pages using slug-only filenames in the root directory with the 'page' layout.


parkr:
@jekyllbot: merge +minor

PR #305, opened 2017-04-19
WordPress: fix variable name for `more_anchor`

dannyvankooten:
In the WordPress importer, the `more_link` variable was never used so this option wasn't really doing anything (always passing `nil`).


parkr:
@jekyllbot: merge +bug

PR #306, opened 2017-04-27
RSS: add "tag" option to add a tag to all posts

sergiocarracedo:
We need to tag posts from RSS source with same tag to differentiate other sources.
And add de TAG option in the RSS importer to set the tag in post header.
If you leave empty no post tagging


.gitignore
```diff
@@ -21,3 +21,5 @@ doc/
 
 .DS_Store
 Gemfile.lock
+
+.idea/
```
pathawks: Might be best to add this to your global gitignore

sergiocarracedo: Done!

parkr:
@jekyllbot: merge +minor

PR #307, opened 2017-05-02
feat: allow additional options to RSS import

dprockcreative:
Allowing additional options to be passed in: "frontmatter" and "body" to define
what is allowed in the frontmatter block (accessible via "page" object in templates)
and which item property/ies will be used for body output


DirtyF:
@dprockcreative Sorry for the delay, could you resolve the conflicts please?

DirtyF:
@jekyllbot: merge +minor

PR #255, opened 2016-05-25
joomla: use & require mysql2 instead of mysql

alejandroerickson:
joomla.rb was using Sequel.mysql, which appears to depend on the gem mysql.  This is depracated and replacing mysql with mysql2 works fine;  it is already in use for wordpress.rb.  I also added the requirement for the mysql2 gem, as this doesn't appear to work without that.

Details of my error message when running the vanilla joomla importer without mysql:

```
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require': LoadError: cannot load such file -- mysql (Sequel::AdapterNotFound)
        from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/adapters/mysql.rb:6:in `rescue in <top (required)>'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/adapters/mysql.rb:3:in `<top (required)>'
        from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/database/connecting.rb:98:in `load_adapter'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/database/connecting.rb:28:in `adapter_class'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/database/connecting.rb:56:in `connect'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/core.rb:108:in `connect'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/core.rb:377:in `adapter_method'
        from /Library/Ruby/Gems/2.0.0/gems/sequel-4.34.0/lib/sequel/core.rb:387:in `mysql'
[...]

PR #309, opened 2017-05-24
Use mysql2 instead of mysql in Joomla3 importer

ienev:
Use and require mysql2 gem instead of mysql for the Joomla3 importer. Change similar to the one for the Joomla importer - #255.

### DIFF
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -48,7 +48,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency('sequel', "~> 3.42")
   s.add_development_dependency('htmlentities', "~> 4.3")
   s.add_development_dependency('hpricot', "~> 0.8")
-  s.add_development_dependency('mysql', "~> 2.8")
   s.add_development_dependency('pg', "~> 0.12")
   s.add_development_dependency('mysql2', "~> 0.3")
   s.add_development_dependency('behance', "~> 0.3")
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -50,6 +50,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('hpricot', "~> 0.8")
   s.add_development_dependency('pg', "~> 0.12")
   s.add_development_dependency('mysql2', "~> 0.3")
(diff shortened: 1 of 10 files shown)


ienev:
AFAIK the `mysql` [gem](https://rubygems.org/gems/mysql/) won't compile with ruby 2.4  due to the Fixnum and Bignum unification:

```shell
compiling mysql.c
mysql.c: In function ‘stmt_bind_result’:
mysql.c:1320:74: error: ‘rb_cFixnum’ undeclared (first use in this function)
  else if (argv[i] == rb_cNumeric || argv[i] == rb_cInteger || argv[i] == rb_cFixnum)
                                                                          ^
mysql.c:1320:74: note: each undeclared identifier is reported only once for each function it appears in
mysql.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-self-assign’
cc1: warning: unrecognized command line option ‘-Wno-constant-logical-operand’
cc1: warning: unrecognized command line option ‘-Wno-parentheses-equality’
cc1: warning: unrecognized command line option ‘-Wno-tautological-compare’
Makefile:241: recipe for target 'mysql.o' failed
make: *** [mysql.o] Error 1

make failed, exit code 2
``` 

The maintainer has deprecated development of the gem and no new version has been released for the past 4 years.

Since there are other importers using `mysql`, may be it is worth applying this change to them too? The gem is also required as a development dependency in the [gemspec](https://github.com/jekyll/jekyll-import/blob/master/jekyll-import.gemspec#L51). This seems like a something that needs to be discussed though (probably in a separate Issue) as it only affects ruby 2.4.

parkr:
> may be it is worth applying this change to them too? 

That sounds great! Want to upgrade all of them in this PR? 😍

ienev:
> Want to upgrade all of them in this PR?

Sure. What about the `.gemspec`? I guess requiring `mysql` won't be needed anymore?

parkr:
> Sure. What about the .gemspec? I guess requiring mysql won't be needed anymore?

Yes! You can replace that with mysql2 or just remove it.

parkr:
@jekyllbot: merge

PR #310, opened 2017-05-24
Add option to specify the MySQL port for the Joomla importers

ienev:
* Add option to specify the MySQL port for the Joomla importers
* Use Ruby 1.9 hash syntax
* Update the Joomla importers documentation

### DIFF
--- a/lib/jekyll-import/importers/joomla.rb
+++ b/lib/jekyll-import/importers/joomla.rb
@@ -14,6 +14,7 @@ module JekyllImport
         c.option 'user', '--user', 'Database user name'
         c.option 'password', '--password', "Database user's password (default: '')"
         c.option 'host', '--host', 'Database host name'
+        c.option 'port', '--port', 'Database port'
         c.option 'section', '--section', 'Table prefix name'
         c.option 'prefix', '--prefix', 'Table prefix name'
       end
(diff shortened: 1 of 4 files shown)


ienev:
@parkr I'll fix the conflicts, but are OK with this PR?

ienev:
Conflicts solved.

lib/jekyll-import/importers/joomla.rb
```diff
@@ -32,11 +33,12 @@ def self.process(options)
         dbname  = options.fetch('dbname')
         user    = options.fetch('user')
         pass    = options.fetch('password', '')
-        host    = options.fetch('host', "localhost")
+        host    = options.fetch('host', 'localhost')
+        port    = options.fetch('port', '3306')
```
parkr: Isn't this port supposed to be an integer?

ienev: Hmm, looks [like](https://github.com/brianmario/mysql2/blob/master/lib/mysql2/client.rb#L85) [it](https://github.com/brianmario/mysql2/blob/master/ext/mysql2/client.c#L78). Should I add `to_i` or change the default value... or both?

parkr: Both! And please update the documentation so it doesn't have the quotes.

ienev:
One of the build jobs seems stuck. Is this normal, should I just wait or there is a way to restart the jobs?

parkr:
No, that's not normal! I tried to restart it but it didn't work. Weird! Merging anyway.

@jekyllbot: merge

PR #311, opened 2017-06-22
add port options for wordpress mysql db connection

passos:


mattr-:
@jekyllbot: merge

lib/jekyll-import/importers/wordpress.rb
```diff
@@ -77,6 +79,7 @@ def self.process(opts)
           :user           => opts.fetch('user', ''),
           :pass           => opts.fetch('password', ''),
           :host           => opts.fetch('host', 'localhost'),
+          :port           => opts.fetch('port', '3306'),
```
parkr: Should this be a string 3306, or an integer 3306?

parkr:
The documentation also needs to be updated.

lib/jekyll-import/importers/wordpress.rb
mattr-: The `mysql2` gem will handle both an integer and a string so it doesn't matter.

PR #312, opened 2017-06-28
Add the new port setting to the Wordpress importer docs

mattr-:
Missed this before when reviewing and merging #311


parkr:
@jekyllbot: merge +docs

Issue #334, opened 2017-12-25
Error when importing from drupal 7


ccamara:
Disclaimer: I am a newbie to jekyll (I've used in a couple of simple sites) and to ruby. 

I wanted to migrate from a drupal 7 site and haven't been able to do it despite I think I've been following the instructions.

That's what I've done so far:

1. Create a blank jekyll site (`jekyll new myblog`)
2. Navigate to `myblog` site.
3. Run the following code:
```

$ ruby -rubygems -e 'require "jekyll-import";
    JekyllImport::Importers::Drupal7.run({
      "dbname"   => "name",
      "user"     => "myuser",
      "password" => "mypassword",
      "host"     => "myhost",
      "prefix"   => "mytableprefix",
      "types"    => ["blog", "post"]
    })'
```
(obviously I replaced my database name, user, password, host, prefix and types).

And that's what I get:
```

Configuration file: /www/jekyll-test/myblog/_config.yml
/var/lib/gems/2.3.0/gems/jekyll-import-0.13.0/lib/jekyll-import/importers/drupal_common.rb:90:in `block (2 levels) in process': undefined method `force_encoding' for []:Array (NoMethodError)
        from /var/lib/gems/2.3.0/gems/jekyll-import-0.13.0/lib/jekyll-import/importers/drupal_common.rb:89:in `each_pair'
        from /var/lib/gems/2.3.0/gems/jekyll-import-0.13.0/lib/jekyll-import/importers/drupal_common.rb:89:in `block in process'
        from /var/lib/gems/2.3.0/gems/sequel-5.3.0/lib/sequel/dataset/actions.rb:151:in `block in each'
        from /var/lib/gems/2.3.0/gems/sequel-5.3.0/lib/sequel/adapters/mysql2.rb:238:in `block (2 levels) in fetch_rows'
        from /var/lib/gems/2.3.0/gems/sequel-5.3.0/lib/sequel/adapters/mysql2.rb:238:in `each'
        from /var/lib/gems/2.3.0/gems/sequel-5.3.0/lib/sequel/adapters/mysql2.rb:238:in `block in fetch_rows'
        from /var/lib/gems/2.3.0/gems/sequel-5.3.0/lib/sequel/adapters/mysql2.rb:152:in `_execute'
        from /var/lib/gems/2.3.0/gems/sequel-5.3.0/lib/sequel/adapters/utils/mysql_mysql2.rb:38:in `block in execute'
[...]

PR #331, opened 2017-12-12
Add support for Drupal 7 installations with PostgreSQL database

kesara:
* Support Drupal 7 installations with PostgreSQL database.
* Updated documentation with an example.
* Minor bug fixes.

PR #339, opened 2018-01-08
Parenthesize the class type

rickysarraf:
Parenthesizing the class is needed for evaluating is_a? on a type
See the closing bug report for full root cause

Closes: https://github.com/jekyll/jekyll-import/issues/334


parkr:
Thanks! This was taken care of in #331 😄

PR #343, opened 2018-02-20
Fixes missing site prefix in user table

Sotilrac:
Fixes missing site prefix in user table in the WordPress Importer. Tested on 4.9.4 WP DB.


DirtyF:
@jekyllbot: :shipit: +minor

DirtyF:
Thanks @Sotilrac 🥇

PR #347, opened 2018-03-01
Use mysql by default in Drupal importers.

argilo:
#331 added support for PostgreSQL in the Drupal importers. According to the documentation, MySQL is supposed to be the default engine, but the `engine` parameter is now mandatory because the default value is not used. This PR fixes that by using the value that was added to the `DEFAULTS` hash.

/cc @kesara


pathawks:
@jekyllbot: :ship: +bug

PR #349, opened 2018-04-03
Added option to configure port. Added missing prefix to taxonomy tables.

dtanzer:

PR #352, opened 2018-04-26
fix when prefix is set.

kristvda:
If a prefix is set, it should also be used by the query to get the tag_group.


DirtyF:
Duplicate of #349

PR #358, opened 2018-10-11
Replace uninitialized constant with alternative

ashmaroli:
`$INPUT_RECORD_SEPARATOR` requires loading `"english"` from Ruby stdlib to initialize it.
So, its better remove the warning from Ruby by using the pre-initialized alternative `$/` especially since Rubocop does not complain about it now..


DirtyF:
@jekyllbot: merge +dev

PR #359, opened 2018-10-11
Remove cyclic require of import command

ashmaroli:
```
require "jekyll-import" > require "jekyll/commands/import" > require "jekyll-import" ...
```


DirtyF:
@jekylbot: merge +dev

PR #360, opened 2018-10-11
fix: Ignore Lint/Loop for tumblr

DirtyF:
Last Rubocop batch broke tumblr importer, this reverts the change and ignore `Lint/Loop` cop for now.


DirtyF:
@jekyllbot: merge +fix

PR #366, opened 2018-10-16
Update Documentation for Contributing to project

ashmaroli:
Update the document to enable easier contribution from the community.


ashmaroli:
> What about writing some tests?

I did consider that.
But since there *isn't a solid pattern* in existing test classes, I won't be able to *stub a test template* reliably..

DirtyF:
@merge +docs

PR #367, opened 2018-10-16
Replace syntax highlighting with a better CSS set

ashmaroli:
Based on https://github.com/jwarby/jekyll-pygments-themes/blob/master/monokai.css
Demo src: https://import.jekyllrb.com/docs/contributing/#creating-a-new-importer

### Before:
before change

### After:
after change


ashmaroli:
> a new Jekyll identity

Oh.. That's news to me!
This can wait until there's a consensus on that then..

PR #369, opened 2018-10-22
fix: duplicate frozen string to avoid error

### LINKED ISSUE #368: Error in WordPress importer
Getting the error `can't modify frozen String` from one of the SQL queries in the importer preventing me from using the importer entirely.

```
ruby -r rubygems -e 'require "jekyll-import"; JekyllImport::Importers::WordPress.run({"dbname"=>"***","user"=>"***","password"=>"***"})'
```

```
Traceback (most recent call last):
        2: from -e:1:in `<main>'
        1: from /var/lib/gems/2.5.0/gems/jekyll-import-0.15.0/lib/jekyll-import/importer.rb:25:in `run'
/var/lib/gems/2.5.0/gems/jekyll-import-0.15.0/lib/jekyll-import/importers/wordpress.rb:166:in `process': can't modify frozen String (FrozenError)
```

https://github.com/jekyll/jekyll-import/blob/e0783f185f4dd73c83a72fe6e8f7094fdaa8c47c/lib/jekyll-import/importers/wordpress.rb#L166

```
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu]
```

DirtyF:
Fix #368

### DIFF
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -12,25 +12,25 @@ module JekyllImport
         types = types.join("' OR n.type = '")
         types = "n.type = '#{types}'"
 
-        query = <<SQL
-                SELECT n.nid,
-                       n.title,
-                       nr.body,
-                       nr.teaser,
-                       n.created,
-                       n.status,
-                       n.type,
-                       GROUP_CONCAT( td.name SEPARATOR '|' ) AS 'tags'
-                FROM #{prefix}node_revisions AS nr,
-                     #{prefix}node AS n
-                     LEFT OUTER JOIN #{prefix}term_node AS tn ON tn.nid = n.nid
-                     LEFT OUTER JOIN #{prefix}term_data AS td ON tn.tid = td.tid
-                WHERE (#{types})
-                  AND n.vid = nr.vid
-                GROUP BY n.nid
-SQL
+        QUERY <<~SQL
+          SELECT n.nid,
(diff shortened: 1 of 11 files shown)


lib/jekyll-import/importers/drupal6.rb
```diff
@@ -12,25 +12,25 @@ def self.build_query(prefix, types, _engine)
...
-                     #{prefix}node AS n
-                     LEFT OUTER JOIN #{prefix}term_node AS tn ON tn.nid = n.nid
-                     LEFT OUTER JOIN #{prefix}term_data AS td ON tn.tid = td.tid
-                WHERE (#{types})
-                  AND n.vid = nr.vid
-                GROUP BY n.nid
-SQL
-
-        query
+        QUERY <<~SQL
```
ashmaroli: Why has this been changed from a local variable `query` to a constant `QUERY` defined within a method..?

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -12,25 +12,25 @@ def self.build_query(prefix, types, _engine)
...
+          SELECT n.nid,
+                 n.title,
+                 nr.body,
+                 nr.teaser,
+                 n.created,
+                 n.status,
+                 n.type,
+                 GROUP_CONCAT( td.name SEPARATOR '|' ) AS tags
```
ashmaroli: `tags` was earlier as `'tags'`..

PR #378, opened 2018-12-11
remove extra / if passed with --url

yyyc514:


ashmaroli:
@yyyc514 I understand the reason behind this.
Since `chomp!` will strip only the trailing slash once, `--url=http://myblog.tumblr.com//` will not be *cleansed* as expected.

The *simplest solution* for the moment is to simply delegate to Ruby's `File` class:
```ruby
url = File.join(url, "/api/read/json/")
```

Additionally, it'd be better if you can add a test for this as well..
Thank you.

yyyc514:
HAHA, that's a typo vs an actual expected mistake, but I hear you.  Might take me a day or two to circle back but sure we can use File.join and I'll see about a test also.

ashmaroli:
@yyyc514 Yes, it is *much better* now. Thank you.

Technically, your `api_feed_url` method has not been declared as a `private` method. But your tests indicate that you intended it to be a `private_class_method` nonetheless.

So I suggest that you declare it explicitly:
```ruby
class << self
  # existing methods in this class

  private # declaration towards the end of the class

  def api_feed_url(url, page, per_page: 50)
    url = File.join(url, "/api/read/json/")
    "#{url}?num=#{per_page}&start=#{page * per_page}"
  end
end
```
*The existing methods cannot be declared as `private` in this PR..*

Additionally, I requested the use of `File.join` due to its simplicity, in comparison to the use of `URI.join` or `Addressable::URI.join`. Feel free to use those if they are more intuitive and better in the current context.

ashmaroli:
> Why are we using `send` here?

Because at the time when the test [was authored](https://github.com/jekyll/jekyll-import/commit/28f0cad7ca48d7636a60b380e94f11d761ac4eb6#diff-39aa2b7bea9e0b946fb976930e39a75bR68), the method was indeed `private`

But `private` became redundant [in this change](https://github.com/jekyll/jekyll-import/commit/eb7f5478367bbb9b324c072b4c3069eb0009d790) and then eventually got removed automatically by RuboCop in subsequent linting sessions because it was "redundant"...

:man_shrugging:

yyyc514:
It's redundant because it was outside the `class < self`, yes?  Those methods should still be private as far as I can tell.

yyyc514:
>  But hiding public methods as private will be a regression (breaking-change).

Ugh.  Indeed. 😐

yyyc514:
Finally. 😇

DirtyF:
@jekyllbot: merge +minor

PR #389, opened 2018-12-23
Replace URI.parse.open with OpenURI.open_uri

### LINKED ISSUE #387: wordpress import error: wrong number of arguments
Hi! Trying to import wordpress xml with: 

ruby -r rubygems -e 'require "jekyll-import";
    JekyllImport::Importers::WordpressDotCom.run({
      "source" => "wordpress.xml",
      "no_fetch_images" => false,
      "assets_folder" => "assets"
    })'

get error: 
Error: wrong number of arguments (given 2, expected 1) 
/usr/lib/ruby/2.5.0/uri/common.rb:236:in `parse'
/home/<user>/.gems/gems/jekyll-import-0.17.0/lib/jekyll-import/importers/wordpressdotcom.rb:41:in `block in download_images'

when I remove the "allow_redirections" argument from line 41 in wordpressdotcom.rb, everything appears to work.

ashmaroli:
`URI.parse` takes only one argument unlike `Kernel#open` as evident from the original `:open` replaced in https://github.com/jekyll/jekyll-import/commit/a23329709018deec7d821c2b23daef68d970e090. The `:allow_redirections` option is made available by the `open_uri_redirections` gem.

Replacing `URI.parse.open` here with a heavy-duty `OpenURI.open_uri` to allow reading the given uri along with `:allow_redirections` option and avoid raising a security flag from RuboCop

Resolves #387

### DIFF
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -38,7 +38,7 @@ module JekyllImport
             next
           end
           begin
-            URI.parse(uri, :allow_redirections => :safe).open do |f|
+            OpenURI.open_uri(uri, :allow_redirections => :safe) do |f|
               File.open(dst, "wb") do |out|
                 out.puts f.read
               end


ashmaroli:
@jekyllbot: merge +bug

PR #393, opened 2019-01-04
drupal6: retain url_alias.dst as permalink in imported nodes

fl3a:
Hi there,
i think, it would be great to retain the Drupal url_alias as permalink in imported nodes.

so i...

- extended the query to be able to fetch url_alias.src

-  assigned permalink within Front Matter

Regards
Florian

### DIFF
--- a/lib/jekyll-import/importers/drupal6.rb
+++ b/lib/jekyll-import/importers/drupal6.rb
@@ -47,9 +47,8 @@ SQL
         data = {
           "excerpt"    => summary,
           "categories" => tags.split("|").uniq,
-          "permalink"  => permalink,
         }
-        
+
         if sql_post_data[:alias]
           data.merge!(
             "permalink" => "/" + sql_post_data[:alias]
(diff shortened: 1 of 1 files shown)


DirtyF (dismissed):
Some style issues left, you can launch tests locally with `script/cibuild`

fl3a:
Fixed all script/cibuild code style offenses.

lib/jekyll-import/importers/drupal6.rb
```diff
@@ -41,10 +43,12 @@ def self.post_data(sql_post_data)
         content = sql_post_data[:body].to_s
         summary = sql_post_data[:teaser].to_s
         tags = (sql_post_data[:tags] || "").downcase.strip
+        permalink = "/" + sql_post_data[:alias]
```
parkr: This should new conditional, right? In case the data `sql_post_data[:alias]` isn't there?

parkr (approved):

parkr:
@jekyllbot: merge +minor

PR #395, opened 2019-01-21
Import Wordpress permalinks

reconbot:
because a good link is forever


lib/jekyll-import/importers/wordpressdotcom.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'uri'
```
ashmaroli: Importers load dependencies via the `self.require_deps` method.
We've already listed `open-uri` to be loaded at runtime. Would that suffice..?

lib/jekyll-import/importers/wordpressdotcom.rb
```diff
@@ -74,6 +75,12 @@ def permalink_title
                                end
         end
 
+        def permalink
+          # Hpricot thinks "link" is a self closing tag so it puts the text of the link after open tag
+          uri = text_for('link') || @node.at('link').following[0].to_s
+          @permalink ||= URI(uri).path
```
ashmaroli: When you memoize like this, every call to method `:permalink` resets variable `uri` unnecessarily.
You should either not memoize this method, or memoize early to assign the variable `uri` just once:
```ruby
def permalink
  @permalink ||= begin
    uri = text_for('link') || @node.at('link').following[0].to_s
    URI(uri).path
  end
end
```

lib/jekyll-import/importers/wordpressdotcom.rb
reconbot: uri is a builtin module for parsing uris, open-uri is for downloading I think? I moved the require

lib/jekyll-import/importers/wordpressdotcom.rb
reconbot: fixed

lib/jekyll-import/importers/wordpressdotcom.rb
ashmaroli: What I meant was, you need not explicitly `require 'uri'` since we're already loading `open-uri`.
I just checked to confirm.
You can call `URI()` without explicitly `require-ing 'uri'`

ashmaroli:
@reconbot Thank you for considering my requests. One minor thing left though.
Our linter script has to pass as well.

Please run `bash script/fmt -a` to fix them.

ashmaroli:
I'm seeing a failure in the test logs.. Do the tests pass locally for you..?
```
Failure: test: WordpressDotComPublishedItem should extract the link as a permalink. (TestWordpressDotComPublishedItem)
```

PR #401, opened 2019-02-15
Fix WP pre tag util raise error when value includes regexp

kinoppyd:
Fix Util#wpautop method to escape backslash.

`String#gsub!` requires escape backslash if second argument as String.
https://docs.ruby-lang.org/en/2.6.0/String.html#method-i-gsub


mattr-:
Thank you for contributing!

@jekyllbot: merge +bug

PR #403, opened 2019-03-09
Add install additional gems note

akshatkedia:
Add note to install required additional gems.


DirtyF:
Thanks! 

@jekyllbot: merge +docs

PR #404, opened 2019-03-17
Create drupal8.rb

vegantriathlete:
I have not tested this since I don't know how to pull this importer in if it's not part of the repository. This should at least get us started.


vegantriathlete:
I have found a way to hack around and test things locally and I'm finding bugs.

PR #410, opened 2019-04-29
Slugify post name with Jekyll::Utils.slugify

ashmaroli:
Delegate to utility function in Core (`Jekyll::Utils.slugify` understands `mode: 'latin'` from Jekyll 3.7.0)


DirtyF:
@jekyllbot: merge +minor

PR #411, opened 2019-04-29
Check whether option["tag"] is null or empty value

ashmaroli:
### Rationale

`option` is expected to be a Hash. Therefore `option.to_s` is never going to be empty since even `{}.to_s` equals `"{}"`. Tracing the git history of the line of code shows that this is probably an error during merge-resolution.
I'm opting to use the more verbose `options["tag"].nil? || options["tag"].empty?` instead of `options["tag"].to_s.empty?` since this conditional is inside an `#each` block. Using the latter format would result in unnecessary allocations of `""` when `options["tag"] == nil`

### DIFF
--- a/lib/jekyll-import/importers/rss.rb
+++ b/lib/jekyll-import/importers/rss.rb
@@ -51,7 +51,7 @@ module JekyllImport
             "title"  => item.title,
           }
 
-          header["tag"] = options["tag"] unless options.to_s.empty?
+          header["tag"] = options["tag"] unless (options["tag"].nil? || options["tag"].empty?)
 
           frontmatter.each do |value|
             header[value] = item.send(value)
--- a/lib/jekyll-import/importers/rss.rb
+++ b/lib/jekyll-import/importers/rss.rb
@@ -51,7 +51,7 @@ module JekyllImport
             "title"  => item.title,
           }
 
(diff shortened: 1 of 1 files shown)


lib/jekyll-import/importers/rss.rb
```diff
@@ -51,7 +51,7 @@ def self.process(options)
             "title"  => item.title,
           }
 
-          header["tag"] = options["tag"] unless options.to_s.empty?
+          header["tag"] = options["tag"] unless (options["tag"].nil? || options["tag"].empty?)
```
mattr-:
```suggestion
          header["tag"] = options["tag"] unless options.fetch("tag", {}).empty?
```

ashmaroli: @mattr- I didn't use `#fetch` because it only checks if `options.key?("tag")`. It will lead to a `nil` assignment when `options["tag"] = nil` is set explicitly.
The chances of the explicit *reset* are slim, albeit very much possible. I didn't want to divert way too much from the existing behavior.

Either ways, I'll go with your call.

mattr-: ah, good to know. let's keep it the way it is then.

DirtyF:
@jekyllbot; merge +fix

Issue #412, opened 2019-05-25
Check that my own importer works locally, on my machine


nicktmro:
Hi all,

I am not a ruby dev so please keep that in mind.

TL;DR
How can I check that a plugin I wrote actually does what I want it to do? How can invoke or run in Terminal / Console the jekyll-import and pass as a parameter my new importer?

Details
I want to extend the behaviour of the RSS importer. I manually edited the `/Library/Ruby/Gems/2.3.0/gems/jekyll-import-0.18.1/lib/jekyll-import/importers/rss.rb` file and saved my changes and it definitely does what I want it to do.

Now this is obviously more than hacky, so I'd like to create a plugin that does what I want it to do (rss_podcast). I followed the instructions on the `Contributing` page on the jekyll-import website but I cannot find a `script/console` command anywhere to check that the plugin I wrote actually works alongside the other plugins.

Thanks a bunch!

ashmaroli:
Couple of stuff:
- **Never make changes to files within an existing gem.** Its not easy to keep track of the bugs you may introduce and forget.
- Clone this repository and make your changes in that clone.
- There is no `script/console`. (The documentation needs updating..)
- Since your importer is based on an existing importer, you can simply *inherit your importer* from the existing one:
  ```ruby
  # New importer in a new file:
  # lib/jekyll-import/importers/rss_podcast.rb
  module JekyllImport
    module Importers
      class RSSPodcast < RSS
        # add just those methods (entire definition) that are being changed
      end
    end
  end
  ```
- Then test it as documented at https://import.jekyllrb.com/docs/rss/ replacing the RSS Importer with yours.

nicktmro:
Thanks for the reply, @ashmaroli. I agree that what I did is super bad, which is why I came to ask for help.

Can you please describe / share the command I need to issue in order to run my extension of the importer? 

Like I said, I'm not a ruby dev, but I thought that running the command from the link you shared would ignore my local changes and would not fail to find the `rss_podcast.rb` file in the path?

Basically, what do I need to run instead of this (how can i specify the path to my local implementation?):

```
$ ruby -r rubygems -e 'require "jekyll-import";
    JekyllImport::Importers::RSSPodcast.run({
      "source" => "my_file.xml"
    })'
```

ashmaroli:
Running Ruby with `-I` will let you set the `$LOAD_PATH`:
```
$ ruby -I lib -r rubygems -e 'require "jekyll-import";
[...]

PR #413, opened 2019-05-31
Add support for Podcast RSS feed imports

nicktmro:
This is very similar to the existing RSS importer. Unlike the RSS importer, this importer will also extract the `enclosure` tag and wrap the url into an `audio` HTML 5 tag so that compliant browsers can just play the audio file.

Standard "shownotes" can be extracted from a feed by specifying the `body` launch argument as per the included documentation.

There is also support to not overwrite previous files using the `overwrite` launch argument with a `false` value. The default is to overwrite any existing files.


nicktmro:
Done

nicktmro:
@ashmaroli I tried extending the RSS Importer (as per your instructions) and leaving just the `def self.process(options)`, since it's where all my changes went, but I get this:
```
rss_podcast.rb:5:in `<module:Importers>': uninitialized constant JekyllImport::Importers::RSS (NameError)
```

As I mentioned in #412 I am not a Ruby dev. My attempts to import / reference the RSS.rb file have all failed, and when I checked all the other Importers, I noticed that none of them actually extend anything other than `Importer` so I coudn't follow a pattern.  My googling is also taking me down rabbit holes and I was wondering if you could shed more light on how to solve the error above?

ashmaroli:
@nicktmro Perhaps the error you're seeing is because the `rss_podcast.rb` is being run before `rss.rb` has been loaded into memory.
So lets try an ensure that the `rss.rb` has been loaded beforehand by adding a call to `require_relative` at the top of the `rss_podcast.rb` file:
```ruby
# frozen_string_literal: true

require_relative "rss"

module JekyllImport
  module Importers
    class RSSPodcast < RSS
      # new definitions
    end
  end
end
```
If the above works, do make the change.
Otherwise, let me know and I shall remove the request to make changes to this PR.
Thank you.

nicktmro:
That did the trick :)

lib/jekyll-import/importers/rss_podcast.rb
```diff
@@ -0,0 +1,68 @@
...
+      # Returns nothing.
+      def self.process(options)
+        source = options.fetch("source")
+        frontmatter = options.fetch("frontmatter", [])
+        body = options.fetch("body", ["description"])
+        overwrite = options.fetch("overwrite", true)
+
+        content = ""
+        uri = URI.parse(source)
+        uri.open { |s| content = s.read }
```
ashmaroli: Any reason you opted for `URI.parse(source).open` instead of `open(source)` used in `rss.rb`?

nicktmro:
CI failed when I used the “open” call.
Something to do with Kernel#open being insecure.

lib/jekyll-import/importers/rss_podcast.rb
nicktmro: CI failed when I used the “open” call.
Something to do with Kernel#open being insecure.

ashmaroli:
The sole changes in this importer as compared to the RSS Importer is actually:
```diff
@@ -30,10 +30,11 @@ module JekyllImport
       # Returns nothing.
       def self.process(options)
         source = options.fetch("source")
         frontmatter = options.fetch("frontmatter", [])
         body = options.fetch("body", ["description"])
+        overwrite = options.fetch("overwrite", true)
 
         content = ""
         open(source) { |s| content = s.read }
         rss = ::RSS::Parser.parse(content, false)
 
@@ -42,10 +43,15 @@ module JekyllImport
         rss.items.each do |item|
           formatted_date = item.date.strftime("%Y-%m-%d")
           post_name = Jekyll::Utils.slugify(item.title, :mode => "latin")
           name = "#{formatted_date}-#{post_name}"
 
+          # Skip this file if it already exists and overwrite is turned off
+          next if !overwrite && File.file?("_posts/#{name}.html")
+
+          audio = item.enclosure.url
+
           header = {
             "layout" => "post",
             "title"  => item.title,
           }
 
@@ -67,10 +73,11 @@ module JekyllImport
           FileUtils.mkdir_p("_posts")
 
           File.open("_posts/#{name}.html", "w") do |f|
             f.puts header.to_yaml
             f.puts "---\n\n"
+            f.puts "<audio controls=\"\"><source src=\"#{audio}\" type=\"audio/mpeg\">Your browser does not support the audio element.</audio>"
             f.puts output
           end
         end
       end
     end
```
The rest are identical.. Perhaps we should just "enhance the existing RSS importer" with some `options` instead of extending it with a few minor changes.. :thinking:

nicktmro:
That sounds good to me, but I probably lack the skill to make that happen. Happy to help test it or justify the current “options”, though.

ashmaroli:
@nicktmro I've merged your rss_podcast importer into the existing rss importer with a few changes. (Notably, I removed the `overwrite` configuration since I felt that to be outside the scope of this PR's primary intention.)
However, feel free to make changes to suit your original use-case and yeah, update the documentation once the implementation is stable.

parkr (approved):
:heart:

parkr:
@jekyllbot: merge +minor

PR #415, opened 2019-08-31
Fix link for Posterous importer

MatteoCampinoti94:
Link was set to movabletype.org


DirtyF:
@jekyllbot: merge +docs

PR #417, opened 2019-09-06
Fix wrong option in self.process_tags

isotopp:
The method self.process_tags in s9y_database.rb wrongly checked the option :categories, likely due to a copy and paste error.

The fix is a one-liner


DirtyF:
Thanks!

@jekyllbot: merge +fix

PR #419, opened 2019-09-06
Change data[author] to a string scalar

isotopp:
Using the minima template, data[author] is supposed to be a string scala, but the conversion turns it into a map of displayname, author_login and author_email. author_login and author_email are also stored as separate scalar attributes. The fix turns data[author] into a string scalar field of displayname, fixing the display of the author name in the minima template.


parkr:
There was a push to migrate away from the scalar author field and to a hash. I wonder how feasible that is, for example if it’s not backwards compatible then GitHub Pages can’t upgrade.

ashmaroli:
> if it’s not backwards compatible then GitHub Pages can’t upgrade

@parkr This plugin is not bundled with `github-pages`, but it is a *"blessed gem"* for Jekyll. Is that cause of your concern?

parkr:
@ashmaroli I was more thinking about the minima gem!

ashmaroli:
> I was more thinking about the minima gem!

I see.. But Minima doesn't care about `author` in a document's front matter or data hash, just the `author` in the config file..

parkr (approved):

parkr:
@jekyllbot: merge +fix

PR #430, opened 2020-03-09
Pluxml

Norore:
PluXML is a blog CMS based on XML files instead of a database for articles and comments.
This script will help people to import their articles and drafts in Jekyll, with Nokogiri library.
I found some troubles with my own articles when you have source code in articles with a format similar to liquid, so I also include an option to tell Jekyll to not use the liquid render in exported files.

### DIFF
--- a/docs/_importers/pluxml.md
+++ b/docs/_importers/pluxml.md
@@ -12,8 +12,8 @@ To import your posts and drafts from a PluXML blog, run:
 {% highlight bash %}
 $ ruby -r rubygems -e 'require "jekyll-import";
     JekyllImport::Importers::Pluxml.run({
-      "source" => "/pluxml/data/articles"
-      "layout" => "your_layout"
+      "source" => "/pluxml/data/articles",
+      "layout" => "your_layout",
       "avoid_liquid" => TRUE
     })'
 {% endhighlight %}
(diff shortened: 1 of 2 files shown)


DirtyF:
@jekyllbot: merge +minor

DirtyF:
@Norore Would you mind opening a new PR to fix Rubocop offenses?

DirtyF:
Fixed most of Rubocop offenses in https://github.com/jekyll/jekyll-import/commit/b3b328e9a88a9f1dc063e864935c099b19e57dbd

docs/_importers/pluxml.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+To import your posts and drafts from a PluXML blog, run:
+
+{% highlight bash %}
+$ ruby -r rubygems -e 'require "jekyll-import";
+    JekyllImport::Importers::Pluxml.run({
+      "source" => "/pluxml/data/articles",
+      "layout" => "your_layout",
+      "avoid_liquid" => TRUE
```
parkr: Using `true` here instead of `TRUE` is recommended, just FYI.

```irb
~$ irb
TRUE
>> TRUE
(irb):1: warning: constant ::TRUE is deprecated
=> true
```

DirtyF: @parkr Right, I've changed it to lowercase.

PR #436, opened 2020-05-17
WordpressDotCom: put downloaded assets into /year/month subfolders to reduce clashes

### LINKED ISSUE #435: WordpressDotCom importer overwrites assets
I tried importing my WordpressDotCom website into Jekyll, but I had 350 images get lost/overwritten. This was due to the authoring tool I was using naming all images:  image1, image2, image3.  
e.g.
```
/2020
  /01
    image1.jpg
    image2.jpg
    image3.jpg
  /02
    image1.jpg
    image2.jpg
    image3.jpg
```

but the WordpressDotCom importer, puts all images into /assets which results in them overwriting each other.

I see that there is a similar issue from 6 years ago that went stale #173

DavidBurela:
This PR resolves #435 

A small change to put downloaded assets into /year/month subfolder. This is done by using the date of the post as the subfolder in the assets folder. This keeps the assets chronologically with the post, and reduces clashes as this is the same as how Wordpress store their files.

**Before:**
When I ran the import tool on my blog, I had 350 clashes of images being overwritten.

**After:**
I successfully ran this on my blog https://DavidBurela.wordpress.com, which has 13 years, 230 posts. I only ended up with 5 clashes (which was due to me referencing images in other posts).

*Note:* I have never coded in Ruby before. I naively implemented it with the smallest number of changes that I could to reduce bugs.

### DIFF
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -192,7 +192,6 @@ module JekyllImport
             content = Hpricot(item.text_for("content:encoded"))
             header["excerpt"] = item.excerpt if item.excerpt
 
-
             # Put the images into a /yyyy/mm/ subfolder to reduce clashes
             assets_sub_folder = assets_folder
             if item.published_at
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -173,7 +173,7 @@ module JekyllImport
             value = meta.at("wp:meta_value").inner_text
             metas[key] = value
           end
-          
+
           author_login = item.text_for("dc:creator").strip
 
           header = {
--- a/lib/jekyll-import/importers/wordpressdotcom.rb
+++ b/lib/jekyll-import/importers/wordpressdotcom.rb
@@ -22,7 +22,7 @@ module JekyllImport
       end
 
       # Will modify post DOM tree
-      def self.download_images(title, post_hpricot, assets_folder, published_at)
+      def self.download_images(title, post_hpricot, assets_folder)
         images = (post_hpricot / "img")
         return if images.empty?
 
@@ -30,12 +30,7 @@ module JekyllImport
         images.each do |i|
           uri = i["src"]
 
-          # Put the images into a /yyyy/mm/ subfolder to reduce clashes
-          date_folder = published_at.strftime("/%Y/%m")
-          assets_with_date_folder = format("%s%s", assets_folder, date_folder)
-          FileUtils.mkdir_p assets_with_date_folder
-
-          dst = File.join(assets_with_date_folder, File.basename(uri))
+          dst = File.join(assets_folder, File.basename(uri))
           i["src"] = File.join("{{ site.baseurl }}", dst)
           Jekyll.logger.info uri
           if File.exist?(dst)
@@ -43,6 +38,7 @@ module JekyllImport
             next
           end
           begin
+            FileUtils.mkdir_p assets_folder
             OpenURI.open_uri(uri, :allow_redirections => :safe) do |f|
               File.open(dst, "wb") do |out|
                 out.puts f.read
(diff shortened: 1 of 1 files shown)


lib/jekyll-import/importers/wordpressdotcom.rb
```diff
@@ -22,16 +22,21 @@ def self.specify_options(c)
       end
 
       # Will modify post DOM tree
-      def self.download_images(title, post_hpricot, assets_folder)
+      def self.download_images(title, post_hpricot, assets_folder, published_at)
```
ashmaroli, line 25: This changes the signature of a public method. Will you be able to pass the `published_at` attribute without changing the arity (no. of arguments a method is called with)?

lib/jekyll-import/importers/wordpressdotcom.rb
```diff
@@ -22,16 +22,21 @@ def self.specify_options(c)
...
 
         Jekyll.logger.info "Downloading images for ", title
         images.each do |i|
           uri = i["src"]
 
-          i["src"] = format("{{ site.baseurl }}/%s/%s", assets_folder, File.basename(uri))
-          dst = File.join(assets_folder, File.basename(uri))
+          # Put the images into a /yyyy/mm/ subfolder to reduce clashes
+          datefolder = published_at.strftime("/%Y/%m")
+          assets_withdate_folder = format("%s%s", assets_folder, datefolder)
```
ashmaroli, line 35: `datefolder` and `withdate` aren't exactly valid words, are they?
```suggestion
          date_folder = published_at.strftime("/%Y/%m")
          assets_with_date_folder = format("%s%s", assets_folder, date_folder)
```

lib/jekyll-import/importers/wordpressdotcom.rb
```diff
@@ -22,16 +22,21 @@ def self.specify_options(c)
...
+          FileUtils.mkdir_p assets_withdate_folder
+
+          i["src"] = format("{{ site.baseurl }}/%s/%s", assets_withdate_folder, File.basename(uri))
+          dst = File.join(assets_withdate_folder, File.basename(uri))
```
ashmaroli, line 39: Wouldn't `File.join("{{ site.baseurl }}", dst)` return the same output as ln#38?

lib/jekyll-import/importers/wordpressdotcom.rb
ashmaroli: Do ensure that all occurrences are renamed. I highlighted just a couple of them.

DavidBurela: agree with this. It wasn't consistent with other variables

lib/jekyll-import/importers/wordpressdotcom.rb
DavidBurela: The call to this method on line 199 has been updated with the extra parameter
Unless there is another place I'm not aware of?

lib/jekyll-import/importers/wordpressdotcom.rb
DavidBurela: Your right, that is WAY simpler.  I just slotted my extra variable into the existing code concat, but rewriting it is cleaner. Updated the PR with this

DavidBurela:
Pulled date subfolder logic earlier in execution chain. This keeps method signature intact. Guarded it with an if statement, so it should also be safer than the original.

Tested it on a fresh import of 161 posts with 837 images.

lib/jekyll-import/importers/wordpressdotcom.rb
DavidBurela: Restored the method signature.  moved the logic earlier and guarded it in case there are pages or posts without a publish date

lib/jekyll-import/importers/wordpressdotcom.rb
```diff
@@ -191,7 +192,14 @@ def self.process(options)
...
 
-            download_images(item.title, content, assets_folder) if fetch
+            # Put the images into a /yyyy/mm/ subfolder to reduce clashes
+            assets_sub_folder = assets_folder
+            if item.published_at
+              date_folder = item.published_at.strftime("/%Y/%m")
+              assets_sub_folder = File.join(assets_folder, date_folder)
+            end
+
+            download_images(item.title, content, assets_sub_folder) if fetch
```
ashmaroli, line 202: `date_folder` isn't used elsewhere so we can avoid creating it. But then again, `assets_sub_folder` will be used only `if fetch`. So wrap the whole thing:
```suggestion
            if fetch
              # Put the images into a /yyyy/mm/ subfolder to reduce clashes
              assets_dir_path = if item.published_at
                                  File.join(assets_folder, item.published_at.strftime("/%Y/%m"))
                                else
                                  assets_folder
                                end

              download_images(item.title, content, assets_dir_path)
            end
```

DavidBurela:
Thanks for your patience on this PR. As I mentioned, I've never used Ruby. So your assistance has been appreciated.

ashmaroli:
> As I mentioned, I've never used Ruby.

Yes, that's the reason I walked you through this.

I did not merge this yet because I'm not sure if everyone would want their assets put into this particular subdir path.

ashmaroli:
@jekyllbot: merge +fix

ashmaroli:
Thank you @DavidBurela

soulfulAdam (approved):

soulfulAdam:
.

PR #454, opened 2021-09-17
Drupal aliases now respect the 'source' config

hugopeixoto:
Alias markdown files were being generated in the current directory instead of the directory defined by the "source" configuration option.

This is my first time using jekyll, still trying to figure it out. I'm sorry if I missed something in how this is supposed to work.

### DIFF
--- a/lib/jekyll-import/importers/drupal_common.rb
+++ b/lib/jekyll-import/importers/drupal_common.rb
@@ -68,6 +68,7 @@ module JekyllImport
           src_dir = conf["source"]
 
           dirs = {
+            :_aliases => src_dir,
             :_posts   => File.join(src_dir, "_posts").to_s,
             :_drafts  => File.join(src_dir, "_drafts").to_s,
             :_layouts => Jekyll.sanitized_path(src_dir, conf["layouts_dir"].to_s),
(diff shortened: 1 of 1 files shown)


mattr-:
:wave: Thanks for taking the time to contribute! Are alias markdown files a Drupal thing? I'm not aware of them.

Either way, this looks good to me. I'll wait a few days to hear back from you about alias markdown files before approving and merging, since I'd like to make sure I understand those properly.

hugopeixoto:
Alright, here's a bit of context:

Drupal's pages are called "nodes". You can access them by navigating to `/node/<nodeid>`. You can also define URL aliases, like mapping `/about-us` to `/node/42`. These aliases are stored in the `url_alias` table.

`jekyll-import` already handles these aliases by going through that table and creating markdown files in the root, like `about-us.md`, with a layout (`refresh.html`) that contains a meta http-equiv redirect to the correct post. With this in place, when you access `/about-us`, you get redirected to `/post/2021-09-20-about-us`. It's not exactly the same behavior, but this is what's currently implemented.

What this PR does is to change the directory where these alias files are created. Instead of always creating them on the root directory, it uses the `source` option in the jekyll configuration, like every other page.

mattr- (approved):

mattr-:
Thanks so much for explaining that!

@jekyllbot: merge +minor

PR #468, opened 2022-03-06
Document additional dependencies of importers

### LINKED ISSUE #467: Please add dependencies to Importer docs
I'm new to Ruby and Jekyll, so I'm not as familiar with how importers and such work.

So if there are "Requirements" for an importer to work, PLEASE document them on the docs page and NOT just the run command.

Just spent a few hours trying to get the Ghost importer to run only to find:

1. I needed to also run "gem install jekyll-import"
2. Then it turns out I needed another gem and had to run "gem install sequel"
3. Then I needed sqlite3 -> "gem install sqlite3"

NEVER assume the reader / user has complete knowledge or expect them to know things, please just be clear with install instructions and usage to avoid confusion

After all that, the Ghost importer still didn't work as it appears "out of date" with Ghost 3 / 4 db schema.  Too much time wasted

ashmaroli:
Introduce a rake task to extract list of available importers into a data file that is then consumed during a future `jekyll build`.
This simplifies maintaining consistency between plugin lib and docs.

Closes #467


docs/_layouts/docs.html
```diff
@@ -10,6 +10,20 @@
       <div class="unit four-fifths">
         <article>
           <h1>{{ page.title }}</h1>
+          {% assign importer = page.path | split: "/" | last | split: "." | first %}
+          {% assign importer_deps = site.data.importer_dependencies[importer] %}
+          {% if importer_deps %}
+          <div class="note info">
+            <h5>Additional Dependencies</h5>
```
parkr, line 17: Could you paste a screenshot of how this looks?

docs/_data/importer_dependencies.yml
```diff
@@ -0,0 +1,109 @@
...
+- safe_yaml
+- sequel
+enki:
+- pg
+- sequel
+ghost:
+- safe_yaml
+- sequel
+- sqlite3
+googlereader:
```
parkr, line 41: Since the filename changed for this importer in the docs, the "next" and "previous" links should be updated. Search the site source for "google_reader" and they should show the places it needs to be changed (remove the underscore).

docs/_data/importer_dependencies.yml
```diff
@@ -0,0 +1,109 @@
...
+- behance
+- safe_yaml
+blogger:
+- rexml/document
+- rexml/parsers/streamparser
+- rexml/streamlistener
+- safe_yaml
+dotclear:
+- active_support
+- active_support/core_ext/string/inflections
```
parkr, line 12: Should we normalize this to the gem it comes from? I wonder if there's a way to find the file using Ruby's $LOAD_PATH and normalize on the gem name. For example, this and the line above are both from the gem "activesupport", I believe.

parkr (commented):
This is fantastic!! Thanks @ashmaroli! I have a couple quick comments that I'd love your thoughts on to improve this even further.

docs/_layouts/docs.html
ashmaroli: Viewport screenshort

drupal_importer_deps

docs/_data/importer_dependencies.yml
ashmaroli: The filename changed. But the permalink did not.
So no links broken.

docs/_data/importer_dependencies.yml
ashmaroli: I thought about it but dropped it for a future pull request..

ashmaroli: Awesome suggestion regarding the use of `$LOAD_PATH`.. I did not think of that..

parkr: Gotcha. This doesn't quite answer the question of how to install the necessary dependencies for the average end user who isn't that familiar with ruby. You and I know how to install activesupport properly but they may not. I wonder how much this change helps without a more explicit instruction like:


```
To use this importer, run the following command beforehand:

    gem install activesupport rexml sequel sqlite3
```

It currently reads to me like a maintainer documentation change rather than an end-user documentation change 🤔

ashmaroli: I see.. I didn't add the `gem install` so that users do not spend time installing gems they *may* already have installed..
But yes, your point is valid and makes sense.

ashmaroli:
Screenshot of current renders

jrnl-deps

pluxml-deps

Rakefile
```diff
@@ -145,4 +145,56 @@ namespace :site do
...
+    importers.each do |importer|
+      doc_name = importer.to_s.split("::").last.downcase
+      deps = importer.require_deps - std_lib
+      next if deps.empty?
+
+      deps.map! do |dep|
+        if dep.start_with?("active_support")
+          "activesupport"
+        elsif dep == "net/http"
+          "net-http"
```
parkr, line 181: Isn't this part of stdlib too? 🤔

Rakefile
```diff
@@ -145,4 +145,56 @@ namespace :site do
       abort "You seem to have misplaced your History.markdown file. I can haz?"
     end
   end
+
+  desc "generate importer-dependencies data file"
+  task :generate_dependency_data do
```
parkr, line 150: Should we add this task as a dependency to the site build/generate task if there is one?

docs/_layouts/docs.html
```diff
@@ -10,6 +10,27 @@
...
+            <p>
+              This importer requires the following additional {% if importer_deps.size == 1 %}library{% else %}libraries{% endif %}.
+            </p>
+            <ul>{% for dep in importer_deps %}
+              <li><code>{{ dep }}</code></li>{% endfor %}
```
parkr, line 22: Maybe we can link to the RubyGems.org page for each dependency?

parkr (approved):

Rakefile
ashmaroli: Ah yes it is..
I misread the [release post for Ruby 3.1.0](https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/).

Will update.

ashmaroli:
Latest Render

ghost-deps

parkr (approved):

parkr:
Great work! :shipit:

parkr:
@jekyllbot: merge +docs

PR #489, opened 2022-11-06
Improve RSS importer with canonical_link and extract_tags option

sumanmaity112:


lib/jekyll-import/importers/rss.rb
```diff
@@ -52,14 +55,24 @@ def self.write_rss_item(item, options)
...
         name = "#{formatted_date}-#{post_name}"
         audio = render_audio && item.enclosure.url
+        canonical_link = options.fetch("canonical_link", false)
 
         header = {
           "layout" => "post",
-          "title"  => item.title,
-        }
+          "title" => item.title,
+          "canonical_url" => (canonical_link ? item.link : nil),
```
parkr, line 63: Hm, this could create a large site with a lot of `canonical_url: nil` in them since the key is present. It might be best to follow the below "tag" method and do something like:

```ruby
header["canonical_url"] = item.link if canonical_link
```

docs/_importers/rss.md
```diff
@@ -17,3 +17,10 @@ $ ruby -r rubygems -e 'require "jekyll-import";
 {% endhighlight %}
 
 The `source` field is required and can be either a local file or a remote one.
+Other optional fields are as follows
```
parkr, line 20:
```suggestion
Other optional fields are as follows:
```

docs/_importers/rss.md
```diff
@@ -17,3 +17,10 @@ $ ruby -r rubygems -e 'require "jekyll-import";
...
+* `canonical_link` add canonical to posts. Default value is `false`
```
parkr, line 21:
```suggestion
* `canonical_link` – copy original link as `canonical_url` to post front matter. (default: `false`)
```

lib/jekyll-import/importers/rss.rb
```diff
@@ -89,6 +102,11 @@ def self.write_rss_item(item, options)
           end
 
           f.puts output
+          if canonical_link
+            f.puts <<~HTML
+              <p>Originally posted on <a href="#{item.link}">#{URI.parse(item.link).host}</a>.</p>
```
parkr, line 107: Remove this – the user can add this if they want it since `canonical_url` is in the front matter. It can be easily added to the `post` layout.

docs/_importers/rss.md
```diff
@@ -17,3 +17,10 @@ $ ruby -r rubygems -e 'require "jekyll-import";
...
+* `render_audio` render <audio> element in posts. Default value is `false`
```
parkr, line 22:
```suggestion
* `render_audio` – render `<audio>` element in posts for the enclosure URLs (default: `false`)
```

docs/_importers/rss.md
```diff
@@ -17,3 +17,10 @@ $ ruby -r rubygems -e 'require "jekyll-import";
...
+* `tag` add specific tag to posts
```
parkr, line 23:
```suggestion
* `tag` – add a specific tag to all posts
```

docs/_importers/rss.md
```diff
@@ -17,3 +17,10 @@ $ ruby -r rubygems -e 'require "jekyll-import";
...
+* `extract_tags` extract tags from given key
```
parkr, line 24: How is this one meant to be used? What variable/field on `item`?

```suggestion
* `extract_tags` – copies tags from the given subfield on the RSS `<item>`
```

parkr (commented):
Thanks! A few comments.

lib/jekyll-import/importers/rss.rb
```diff
@@ -52,14 +55,24 @@ def self.write_rss_item(item, options)
...
 
         header["tag"] = options["tag"] unless options["tag"].nil? || options["tag"].empty?
 
+        if options["extract_tags"]
+          tags_from_feed = item.instance_variable_get("@#{options["extract_tags"]}")
+          unless tags_from_feed.nil?
+            tags = tags_from_feed.map { |feed_tag| feed_tag.content.downcase }
+            header["tag"] = tags unless tags.empty?
+          end
+        end
```
ashmaroli, line 74: This section of code is confusing.
- First we check for `options["extract_tags"]` where `options` is the CLI options.
- Then we check for `item.instance_variable_get("@#{options["extract_tags"]}")` using same `options` object. So why not store the `key` in a variable for reuse?
- Then we preprocess feed-tags using `#map`. That means we expect an array of `feed_tag` instead of *any non-nil object*. So, we should explicitly test for array instead. We also *downcase* feed_tags. So, if the feed originally contained `["holiday", "Holiday"]`, we'd end up with: `["holiday", "holiday"]`. So we should filter duplicates.

```suggestion
        tag_name = options["extract_tags"]
        if tag_name
          tags_from_feed = item.instance_variable_get("@#{tag_name}")
          if tags_from_feed.is_a?(Array)
            tags = tags_from_feed.map { |feed_tag| feed_tag.content.downcase }
            tags.uniq!
            header["tag"] = tags unless tags.empty?
          end
        end
```

ashmaroli (commented):
Needs some changes to be made.

lib/jekyll-import/importers/rss.rb
sumanmaity112: sure, will update it

lib/jekyll-import/importers/rss.rb
sumanmaity112: Hi @parkr, that's correct, it can be easily do able by making simple change it `post` layout. But I would like to have it here due to following reasons:
* it's configurable using `--canonical_link` option
* I believe multiple people wants to have reference to the original post. To make every one of them to change `post` layout it's little too much according to me.

 In case if it's not preferred to inter connect frontmatter header and this footer, maybe adding a new option will be better.

lib/jekyll-import/importers/rss.rb
sumanmaity112: Sure, this code looks much better :). Only thing `tag_name` is actually the `key` mentioned by CLI options. Therefor I'll update it accordingly.

sumanmaity112:
I'll make the required/suggested changes and update the same PR

ashmaroli:
@sumanmaity112 This also requires tests especially because the implementation *expects* a particular state. Starting with some basic tests will do.

lib/jekyll-import/importers/rss.rb
ashmaroli: I know that CLI option param `KEY` corresponds to the `tag_name` in my suggestion. But now that you mentioned it, I wonder why the option is in plural `extract_tags` and why not singular as in
`--extract_tag TAG_NAME`..?

sumanmaity112: Hi @ashmaroli, the reason of introducing `--extract_tags` is to extract tag(s) from RSS feed. When someone add a post they'll have multiple tags. In this case, you'll provide the **key**(sub field of RSS payload) and it'll add all those tags to frontmatter. 

For example, if you look into the follow RSS feed from Medium, it has multiple **category** (which is actually tags).

```xml
<?xml version="1.0" encoding="UTF-8"?>
<rss xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:content="http://purl.org/rss/1.0/modules/content/"
     xmlns:atom="http://www.w3.org/2005/Atom" version="2.0"
     xmlns:cc="http://cyber.law.harvard.edu/rss/creativeCommonsRssModule.html">
    <channel>
        <title><![CDATA[Stories by Suman Maity on Medium]]></title>
        <description><![CDATA[Stories by Suman Maity on Medium]]></description>
        <link>https://medium.com/@suman.maity112?source=rss-92eeb3145de9------2</link>
        <image>
            <url>https://cdn-images-1.medium.com/fit/c/150/150/1*rAk1Q7-7kFOM-arRqtpvMA.jpeg</url>
            <title>Stories by Suman Maity on Medium</title>
            <link>https://medium.com/@suman.maity112?source=rss-92eeb3145de9------2</link>
        </image>
        <generator>Medium</generator>
        <lastBuildDate>Mon, 07 Nov 2022 19:38:06 GMT</lastBuildDate>
        <atom:link href="https://medium.com/@suman.maity112/feed" rel="self" type="application/rss+xml"/>
        <webMaster><![CDATA[yourfriends@medium.com]]></webMaster>
        <atom:link href="http://medium.superfeedr.com" rel="hub"/>
        <item>
            <title><![CDATA[Is it the era of mono-repo?]]></title>
            <link>
                https://medium.com/@suman.maity112/is-it-the-era-of-mono-repo-671e6dee387?source=rss-92eeb3145de9------2
            </link>
            <guid isPermaLink="false">https://medium.com/p/671e6dee387</guid>
            <category><![CDATA[version-control]]></category>
            <category><![CDATA[multi-repo]]></category>
            <category><![CDATA[monorepo]]></category>
            <dc:creator><![CDATA[Suman Maity]]></dc:creator>
            <pubDate>Mon, 07 Nov 2022 08:11:46 GMT</pubDate>
            <atom:updated>2022-11-07T08:11:46.975Z</atom:updated>
            <content:encoded><![CDATA[dummy content]]></content:encoded>
        </item>
    </channel>
</rss>```

This is the main reason of having `--extract_tags` as plural instead of singular. You can find the imported RSS feed [here](https://sumanmaity112.github.io/posts/is-it-the-era-of-mono-repo/)

File: lib/jekyll-import/importers/rss.rb (hunk 1 above)
Comment (sumanmaity112) [reply lib/jekyll-import/importers/rss.rb:63]: Hi @parkr, we don't have change it to `item.link if canonical_link` as `.compact` will take care of null values

**Comment by sumanmaity112** [comment]
> @sumanmaity112 This also requires tests especially because the implementation _expects_ a particular state. Starting with some basic tests will do.

Hi @ashmaroli, didn't got your point. Are you asking me to add some sort unit tests?

**Comment by ashmaroli** [comment]
Hello again, @sumanmaity112
There are couple of issues with the new "private" method you have defined:
- In Ruby, `private` keyword only affects the *instance methods*. So, the `self.get_tags` method is still public. You will have to use the `private_class_method :get_tags` declaration to make it truly private.
- Please consider refactoring the `self.get_tags` once more. The following lines do not seem right:
   ```ruby
   explicit_tag = options["tag"]
   if explicit_tag
     return explicit_tag unless explicit_tag.nil? || explicit_tag.empty?
   ```
   `nil` is a falsy. The check for `explicit_tag.nil?` is redundant after the initial `if explicit_tag` check. Plus, what happened to the exclusivity of the two options? Why ignore `options["extract_tags"]` if `explicit_tag` is nil or undefined?
(This is why test coverage is important)

**Comment by sumanmaity112** [comment]
@ashmaroli, refactored the `get_tags` method. Please do have a look and let me know if I have to update something else

File: lib/jekyll-import/importers/rss.rb [hunk 9]
```diff
@@ -89,6 +93,26 @@ def self.write_rss_item(item, options)
...
+        tags_reference = options["extract_tags"]
+        if tags_reference
+          tags_from_feed = item.instance_variable_get("@#{tags_reference}")
+          if tags_from_feed.is_a?(Array)
+            tags = tags_from_feed.map { |feed_tag| feed_tag.content.downcase }
+            tags.uniq!
+            return tags unless tags.empty?
+          end
         end
       end
```
ashmaroli, line 117: Here's how I would implement it doing away with nested conditionals.

```suggestion
      def self.get_tags(item, options)
        explicit_tag = options["tag"]
        return explicit_tag unless explicit_tag.nil? || explicit_tag.empty?

        tags_reference = options["extract_tags"]
        return unless tags_reference

        tags_from_feed = item.instance_variable_get("@#{tags_reference}")
        return unless tags_from_feed.is_a?(Array)

        tags = tags_from_feed.map { |feed_tag| feed_tag.content.downcase }
        tags.empty? ? nil : tags.tap(&:uniq!)
      end
      private_class_method :get_tags
```
We don't have test coverage on this. So, we'll have to take the code at face-value.

sumanmaity112:
@ashmaroli, @parkr  Any idea how to fix **Continuous Integration / Ruby 2.5**? It failing to setup the Ruby 2.5 itself

ashmaroli:
@sumanmaity112 To fix the CI for Ruby 2.5, we will need to lock gem `rdoc` to an older version.
But that is a maintenance task and falls outside the scope of this PR. If you have the time and have the technical know-how regarding this, feel free to submit a pull request. Either ways, I will tackle it in a couple of days.

parkr:
Locked. If you're comfortable rebasing, please do so!

sumanmaity112:
Sure, will rebase the changes

sumanmaity112:
Hi @parkr , I rebased the changes couple days back. Please let me know if anything else is pending from my side which is required to merge this PR.

ashmaroli:
Hello @sumanmaity112,
One final nitpick from my end:

*Could we have the CLI option description and docs mirror each other?*

I don't see why the concise description in the docs wouldn't be apt for the CLI context display as well.

sumanmaity112:
sure, will update it

sumanmaity112:
Hi @parkr, Could you please check if it's better now

ashmaroli (approved):
LGTM!

lib/jekyll-import/importers/rss.rb
```diff
@@ -89,8 +93,28 @@ def self.write_rss_item(item, options)
           end
 
           f.puts output
+          if canonical_link
+            f.puts <<~HTML
+              <p>Originally posted on <a href="#{item.link}">#{URI.parse(item.link).host}</a>.</p>
```
parkr, line 98: I still think we should remove this before merging. What if the blog is not in English? Or if they just don't want this link? Now the user has to manually edit all of their posts. This is asking way more of the user than editing their post layout. In general, injecting our own *content* is not what we do. Metadata is fine but I'm not comfortable writing content to these posts.

> it's configurable using `--canonical_link` option

True but then you lose the front matter value. The docs for this only says you get the front matter value, not extra content. We can trust that the user can add a link themselves if they so desire. Adding canonical_url is sufficient.

sumanmaity112: Hi @parkr, Your point is valid in terms of user can add it if they want. But I am not sure how it can be automated?

For my case, I import all my posts from Medium to github pages. I want to add **canonical_url** at the end of the imported content. You can see the [example here](https://sumanmaity112.github.io/posts/is-it-the-era-of-mono-repo/#conclusion). Currently, I have a [Github Workflow](https://github.com/sumanmaity112/sumanmaity112.github.io/blob/main/.github/workflows/import.yml) to import all posts from Medium. 
At least I am not sure how to achieve this if the importer is not adding it automatically. Maybe you can help me here. Thanks

parkr: Cool workflow! Nice way of having a backup/mirror of your site.

I mean that adding it to the post layout adds this link to every post automatically since every post used the post layout, right? Even if they don't, you could add front matter defaults so that each post uses the post layout, then add the link to the post layout.

1. Post itself contains only the content the user wrote
2. _layouts/post.html contains canonical link if specified
3. Each post specifies that it should use the post layout, or use front matter defaults to specify the post layout should be used by all posts

I don't want this importer to modifying the user's post contents (especially when it's not documented).

sumanmaity112: @parkr Your point is valid.

Now in my mind there are couple of options
1. Remove the changes which adds canonical url to the post
2. Keep the changes and document it
3. Add a new options (addition to `--canonical_link`) to configure this specific footer content

Please let me know whatever you think is better.

ashmaroli: @sumanmaity112 Since Parker doesn't want importers to modify *imported content* regardless of configuration, it's best that you remove the `canonical_url` **markup injection**. Let the option to add the info to front matter remain for subsequent consumption.

Regarding your personal blog, you have two options:
1. Modify the footer template in your theme to render the canonical url.
2. Write a custom Ruby plugin into a `_plugins` directory in your blog source repository. The plugin can easily inject custom content into your posts at runtime via Jekyll Hooks.

parkr (approved):
Thank you!

parkr:
@jekyllbot: merge +minor

PR #491, opened 2022-11-20
Lock `rdoc` gem to v6.3.x to avoid Psych 4.0

ashmaroli:
RDoc v6.4 has Psych v4 as a runtime dependency. Psych v4 breaks our CI running on Ruby 2.5
Therefore, restrain installation to rdoc-6.3.x until we stop testing with Ruby 2.5.

### DIFF
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   # development dependencies
   s.add_development_dependency("bundler")
   s.add_development_dependency("rake", "~> 12.0")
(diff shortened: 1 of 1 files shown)


ashmaroli:
@jekyllbot: merge +dev

parkr:
Let's upgrade CI to remove EOL rubies soon, then we can remove this lock. Thanks for the quick fix, @ashmaroli!

parkr:
#492 tracks removing EOL rubies.

This was indirectly for https://github.com/jekyll/jekyll-import/pull/489.

parkr:
Can we revert this now @ashmaroli ? Since we have now removed 2.5.

PR #493, opened 2022-11-20
Bump dependencies and CI Ruby versions

### LINKED ISSUE #492: Remove EOL Rubies from CI
2.5 is EOL, so we should consider removing explicit support for it (we can leave implicit support for it though)

### LINKED ISSUE #494: Add Ruby 3.x support
Add ruby 3.0 and 3.1 to CI and fix issues such that CI passes.

ashmaroli:
Closes #492
Closes #494

### DIFF
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -41,7 +41,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency("simplecov-gem-adapter", "~> 1.0")
 
   # migrator dependencies:
-  s.add_development_dependency("behance", "~> 0.3")
+  # s.add_development_dependency("behance", "~> 0.3") # uses outdated dependencies
   s.add_development_dependency("hpricot", "~> 0.8")
   s.add_development_dependency("htmlentities", "~> 4.3")
   s.add_development_dependency("mysql2", "~> 0.3")
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -25,18 +25,18 @@ Gem::Specification.new do |s|
   # runtime dependencies
   s.add_runtime_dependency("jekyll", ">= 3.7", "< 5.0")
   s.add_runtime_dependency("nokogiri", "~> 1.0")
-  s.add_runtime_dependency("reverse_markdown", "~> 1.0")
+  s.add_runtime_dependency("reverse_markdown", "~> 2.1")
 
   # development dependencies
   s.add_development_dependency("bundler")
-  s.add_development_dependency("rake", "~> 12.0")
+  s.add_development_dependency("rake", "~> 13.0")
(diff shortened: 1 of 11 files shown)


ashmaroli:
Looks like we have reached a stalemate.. :-(

parkr:
Maybe we need to make this about dropping EOL Rubies and make a separate CL about adding Ruby 3 support. What do you think?

parkr:
Another thing to consider is limiting old importers to a specific ruby version. We can add runtime checks and documentation to say a given importer is limited to Ruby 2.x. It’s not ideal but this gem and its importers are best-effort.

jekyll-import.gemspec
```diff
@@ -25,29 +25,30 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency("rr", "~> 1.0")
   s.add_development_dependency("rubocop-jekyll", "~> 0.11.0")
-  s.add_development_dependency("shoulda", "~> 3.5")
+  s.add_development_dependency("shoulda", "~> 4.0")
   s.add_development_dependency("simplecov", "~> 0.7")
   s.add_development_dependency("simplecov-gem-adapter", "~> 1.0")
 
   # migrator dependencies:
-  s.add_development_dependency("behance", "~> 0.3")
+  # s.add_development_dependency("behance", "~> 0.3") # uses outdated dependencies
```
parkr, line 44: Shall we move this to the Gemfile and add a Ruby version check or just wait until someone comes around and tries to add tests for the Behance importer?

test/helper.rb
```diff
@@ -17,7 +17,7 @@ class Test::Unit::AssertionFailedError < ActiveSupport::TestCase::Assertion
 require File.expand_path("../lib/jekyll-import.rb", __dir__)
 include JekyllImport
 
-JekyllImport::Importer.subclasses.each(&:require_deps)
+# JekyllImport::Importer.subclasses.each(&:require_deps)
```
parkr, line 20: Let's remove this line.

parkr (approved):
Just a couple cleanup comments. Thanks for tackling this!

jekyll-import.gemspec
ashmaroli: We shall wait for someone to add tests for the Behance importer or at minimum report a bug or some other issue with the Behance importer..
I'm not keen on moving dev_deps to Gemfile (for this project) because the Rubygems dev_deps listing may serve as secondary documentation for users.

ashmaroli:
@jekyllbot: merge +dev

jekyll-import.gemspec
parkr: Ok. Commenting this out removes it from the dev deps anyway, so I'm not sure it's of much use here.

PR #496, opened 2022-11-28
Define CLI option descriptions consistently

ashmaroli:
## Summary

- List options with default values after ones that do not.
- Terminate sentences with a period.
- Mention default values with a consistent pattern `(default: ...)`, at the end of description.
- Use double-quotes to delimit strings, optionally single-quotes for substrings.
- Align parameters for better readability for humans (*opinionated.*).


parkr:
I haven't looked through these yet but I was thinking that we could auto-generate the documentation from these options somehow (in a separate PR). Thanks for cleaning these up!

ashmaroli:
> I was thinking that we could auto-generate the documentation from these options

Yes, we could. I have already developed a vision / idea locally.
I found current docs theme off-putting so came up with my own opinionated design to suit the local development.

parkr (approved):

parkr:
@jekyllbot: merge +docs

PR #497, opened 2022-11-28
Add CI workflow job to print importer help info

ashmaroli:
To have a one-stop place to visualize what an end-user would see on invoking the help menu for desired importer.


ashmaroli:
Unfortunately, GHA sanitizes mentions of `--password` to `-***`. So, this idea probably won't succeed.. 😣

parkr:
We could write these to the documentation so that when contributors change the options, the documentation is automatically updated by an Action.

PR #499, opened 2022-11-28
Introduce medium importer

sumanmaity112:


docs/_importers/medium.md
```diff
@@ -0,0 +1,25 @@
...
+permalink: /docs/medium/
+---
+
+To import your posts from [Medium](https://medium.com/), run:
+
+{% highlight bash %}
+$ ruby -r rubygems -e 'require "jekyll-import";
+    JekyllImport::Importers::Medium.run({
+        "username"       => "name",
+        "render_audio"   => "false",
```
parkr, line 16: Same here, is this also meant to be a plain Boolean or a string value?

lib/jekyll-import/importers/medium.rb
```diff
@@ -0,0 +1,32 @@
...
+        c.option "canonical_link", "--canonical_link", "Copy original link as canonical_url to post. (default: false)"
+        c.option "render_audio", "--render_audio", "Render <audio> element in posts for the enclosure URLs (default: false)"
+      end
+
+      def self.validate(options)
+        abort "Missing mandatory option --username." if options["username"].nil?
+      end
+
+      def self.require_deps
+        JekyllImport.require_with_fallback(%w(
```
parkr, line 17: Since we're using RSS, should we can RSS.require_deps instead?

parkr (approved):
Minor comments. LGTM

ashmaroli:
Hello @sumanmaity112,
I am posting a generic comment about my observations instead of a review.

- If you take a look at existing importers other than the RSS importer, you'll see that `:specify_options` definition has a pattern of aligning the various option parameters. I would like you to follow the same alignment style here as well. (In future, I may set up RuboCop to do this for contributors).
- You have `--canonical_link` documented as `false` by default. But the `:process` method has it default to `true`. **This needs to be addressed**.
- I find the description for `--render_audio` going over my head. What do you mean by *enclosure URLs*?
- We have a Rake task to generate dependency metadata of importers for our docs site. You have to run `bundle exec rake site:generate_dependency_data` to update the data store for every new importer (and if dependencies for existing importers change). My apology for not having this step documented in our Contribution guide. *In future, I plan to have Jekyll automate this for us*.
- With the above point in mind, if you choose to  update to using `Importers::RSS.require_deps` as @parkr wondered, you have to run the above rake task once again to see if the task is able to register dependencies correctly. If not revert to the original API and update the list as needed.
- I suggest using `Importers::RSS.process` in `self.process` instead of current `RSS.process` because it took me some time to realise that you were calling the RSS importer instead of the RSS library from the `rss` gem.
- You have set `extract_tags` to invariably extract `category` field info from the RSS feed but have not documented that anywhere. Please leave a Ruby comment above the method for future reference for contributors and maintainers as to why this was done. Additionally, please document a tamer version of the same comment for end-user laymen so that they know what to expect.

docs/_importers/medium.md
```diff
@@ -0,0 +1,25 @@
...
+        "canonical_link" => "true",
```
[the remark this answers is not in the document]
sumanmaity112: Updated the example

docs/_importers/medium.md
sumanmaity112: Updated the example

sumanmaity112:
Hi @ashmaroli 

- I tried to update the `importer_dependencies.yml` using `bundle exec rake site:generate_dependency_data` but it started changing dependencies for other importer as well (without any of my changes). I am not sure if it's expected, so for now I manually updated the dependencies for Medium importer.

- About the descriptions for `--render_audio`, it has the same descriptions as mentioned in RSS importer. Please let me know if I have to change it or not. If we want to change then I'll prefer to keep it sync with RSS importer as medium importer just pass the option value to RSS importer.

parkr:
The render_audio documentation was something I asked to be like that. Before it was too vague. Perhaps now it's too specific. This option is basically only useful if you have a podcast RSS feed I think.

sumanmaity112:
> The render_audio documentation was something I asked to be like that. Before it was too vague. Perhaps now it's too specific. This option is basically only useful if you have a podcast RSS feed I think.

Maybe, I am also not sure. This option was there before my changes in RSS importer.

lib/jekyll-import/importers/medium.rb
```diff
@@ -0,0 +1,31 @@
...
+        Importers::RSS.require_deps
+      end
+
+      def self.process(options)
+        Importers::RSS.process({
+          "source"         => "https://medium.com/feed/@#{options.fetch("username")}",
+          "render_audio"   => options.fetch("render_audio", false),
+          "canonical_link" => options.fetch("canonical_link", false),
+          # Add existing tags from Medium post to front matter using `category` subfield on the RSS <item>
+          "extract_tags"   => "category",
```
parkr, line 26: Out of curiosity, why not support all of the RSS options? Should we just suggest to users on the docs page for Medium to use the RSS importer and show them how to use it?

lib/jekyll-import/importers/medium.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      def self.require_deps
+        Importers::RSS.require_deps
+      end
+
+      def self.process(options)
+        Importers::RSS.process({
+          "source"         => "https://medium.com/feed/@#{options.fetch("username")}",
+          "render_audio"   => options.fetch("render_audio", false),
```
parkr, line 23: Do Medium articles typically have audio?

sumanmaity112: Till now I haven't found any post with audio option. But if I am not wrong, it's possible to add video in medium post

lib/jekyll-import/importers/medium.rb
sumanmaity112: > Out of curiosity, why not support all of the RSS options?

If you see all the options from RSS importer is there except **tag**.

> Should we just suggest to users on the docs page for Medium to use the RSS importer and show them how to use it?

It's possible but it'll be little verbose. Currently this Medium importer is just the wrapper for RSS importer.

ashmaroli:
Thank you for making the changes as suggested, @sumanmaity112. I have pushed a commit to your branch. Do run `git pull origin master` before you make additional changes to this proposal.

ashmaroli:
I have some enquiries remaining @sumanmaity112.
 - Is this importer done being developed? As in, if the RSS importer gets new options, how do we decide if those options need to be ported to the Medium Importer?
 - It is still not clear to me **why** you explicitly chose to extract the `category` field from the feed XML. The Ruby comment you have left needs some more love.
 - Ideally, we ask contributors to attach tests for pull requests. But given the state of the test-suite for this project, that will be overlooked. Can you instead attach a **mock rss file** based on your canonical site's export file? You may attach the mock file in the test directory.

sumanmaity112:
Hi @ashmaroli,
> Is this importer done being developed? As in, if the RSS importer gets new options, how do we decide if those options need to be ported to the Medium Importer?

This is depends on the developer. If a new feature is getting added and same feature can be usable in case of Medium import then it's better to add it. According to me, we should have similar **options** set as **RSS** import. At the end, Medium import is nothing more than a wrapper around RSS importer with some default configurations.

> It is still not clear to me why you explicitly chose to extract the category field from the feed XML. The Ruby comment you have left needs some more love.

I explained the comment little bit more, please do have a look.

> Ideally, we ask contributors to attach tests for pull requests. But given the state of the test-suite for this project, that will be overlooked. Can you instead attach a mock rss file based on your canonical site's export file? You may attach the mock file in the test directory.

I added mock RSS feed file generated from my personal Medium feed (only with couple of items). Please let me know if I have to change anything there.

parkr:
Needs a rebase to get @ashmaroli's latest commit to lock Psych to 4.x. I filed a request with ruby/setup-ruby: https://github.com/ruby/setup-ruby/issues/412

sumanmaity112:
@parkr Rebase is done. Now all CI verification also passed.

parkr:
Thank you!

@jekyllbot: merge +minor

PR #500, opened 2022-11-30
Build and deploy docs site via GitHub Actions

ashmaroli:
Similar to how the docs are deployed for the Jekyll Core repository.

## Rationale

- **Allow building with plugins to inject data at runtime.**
- Allow building with latest version of Jekyll.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,6 +7,9 @@ on:
 
 env:
   RUBY_VERSION: 2.7
+  
+  # For control over build environment
+  DOCS_DEPLOY: true
 
 jobs:
   deploy_docs:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,46 @@
+name: Build and deploy Documentation site
+
+on:
+  push:
+    branches:
+      - master
+
+env:
+  RUBY_VERSION: 2.7
+
+jobs:
+  deploy_docs:
+    runs-on: "ubuntu-latest"
+    env:
+      BUNDLE_PATH: "vendor/bundle"
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,49 @@
...
+          rm -rf docs/_site/
+          git clone --depth=1 --branch="${REMOTE_BRANCH}" --single-branch --no-checkout \
+            "${REMOTE_REPO}" docs/_site/
+      - name: Build site
+        run: bundle exec jekyll build --source docs --destination docs/_site --verbose --trace
+      - name: Deploy to GitHub Pages
+        run: |
+          SOURCE_COMMIT="$(git log -1 --pretty="%an: %B" "$GITHUB_SHA")"
+          pushd docs/_site &>/dev/null
+          : > .nojekyll
```
parkr, line 40: Why not use "touch" here? This command is a bit obtuse.

parkr (approved):

.github/workflows/docs.yml
ashmaroli: `touch <file>` and `: > <file>` is subtly different even though they appear to do the same at first glance.

`touch <file>` creates `<file>` if it doesn't exist and just modifies *last accessed time `atime`* on subsequent uses.

`: > <file>` also creates `<file>` if it doesn't exist but will *truncate its length to `0`* (changes `mtime`) on subsequent uses.

Ref: [Stack Exchange Thread](https://askubuntu.com/questions/831414/what-is-the-difference-between-touch-file-and-file).

---

For argument's sake, one could state that in the current context, the maintainers have total control over `.nojekyll` never being a *non-empty* file and that its contents are irrelevant regardless.

ashmaroli:
@jekyllbot: merge +dev

PR #501, opened 2022-12-01
Omit importer dependencies in docs-site workflow

ashmaroli:
## Summary

Avoid installing importer dependencies and test-suite dependencies while building / deploying the documentation site.

## Background

While setting-up *Deploy Previews* (on `pull_request` events) for the docs-site via Netlify. the build-image there failed to install certain gems with native=extensions due to missing dependencies of the native-code compiler.

## Advantages

- Lesser number of gems to install.
- Using conditional block in the gemspec instead of a Bundler group in the Gemfile means dev-dependencies continue getting listed at the Rubygems.org page as is current behavior.

### DIFF
--- a/jekyll-import.gemspec
+++ b/jekyll-import.gemspec
@@ -32,25 +32,32 @@ Gem::Specification.new do |s|
   s.add_development_dependency("rake", "~> 13.0")
   s.add_development_dependency("rdoc", "~> 6.0")
 
-  # test dependencies:
-  s.add_development_dependency("redgreen", "~> 1.2")
-  s.add_development_dependency("rr", "~> 1.0")
-  s.add_development_dependency("rubocop-jekyll", "~> 0.11.0")
-  s.add_development_dependency("shoulda", "~> 4.0")
-  s.add_development_dependency("simplecov", "~> 0.7")
-  s.add_development_dependency("simplecov-gem-adapter", "~> 1.0")
+  # Dependencies not needed during building / deployment of the documentation site.
+  #
+  # Containing them within a conditional block in the gemspec instead of using a Bundler
+  # group in the Gemfile ensures that they remain listed in the page at Rubygems.org.
 
-  # migrator dependencies:
-  # s.add_development_dependency("behance", "~> 0.3") # uses outdated dependencies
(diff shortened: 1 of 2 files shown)


jekyll-import.gemspec
```diff
@@ -32,25 +32,32 @@ Gem::Specification.new do |s|
...
-  s.add_development_dependency("rss", "~> 0.2")
-  s.add_development_dependency("sequel", "~> 5.62")
-  s.add_development_dependency("sqlite3", "~> 1.3")
-  s.add_development_dependency("unidecode", "~> 1.0")
+  # Dependencies not needed during building / deployment of the documentation site.
+  #
+  # Containing them within a conditional block in the gemspec instead of using a Bundler
+  # group in the Gemfile ensures that they remain listed in the page at Rubygems.org.
+
+  unless ENV["DOCS_DEPLOY"]
```
parkr, line 40: Bundler has first-class support for this kind of thing with groups. Would it make more sense to use groups?

parkr (approved):

jekyll-import.gemspec
ashmaroli: Isn't the Ruby comment on the preceding lines stating exactly *why I did not use a Bundler group* or is the comment not clear enough..?

parkr: I see it now, yes. I have been reviewing on GitHub Mobile which does not show all the context – even less when you're working on a comment. Do you think listing the dev dependencies on RubyGems.org is doing anyone any good? I don't think I have ever looked at dev dependencies before... How would they know which dependencies are for each importer? I feel like folks would actually prefer this all on one page, on our docs site. So should we move these to the Gemfile and just leave it be? Since we're thinking about these things, could we do better documenting dependencies?

In terms of ergonomics, jekyll-import leaves a lot to be desired. The first time you run, you get an error since it's very unlikely you already have all the dependencies. Could we launch bundler and auto-install dependencies somehow? And we even ask users to run the Ruby interpreter and input a Ruby program. We don't even propose users use the CLI...

parkr:
@jekyllbot: merge +dev

PR #503, opened 2022-12-04
New Docs

ashmaroli:


ashmaroli:
@parkr Hello. Even though this is still a WIP, I believe it's developed enough for a *preview* and *first-round* of feedback on the Frontend / UX.
I have set up Netlify to render deploy previews for pull requests. You'll find the links to the preview in a comment at the top of this page 👆
Thanks.

parkr:
Hey @ashmaroli, this is in my queue to review but my high level thoughts are:

1. Lack of consistency with Jekyllrb.com is not desirable IMO. I prefer them to be of the same general structure so that it's easy to navigate between the two and both feel familiar.
1. I really like the new pages for each specific importer, especially using the CLI directly. 
1. I think the massive note about dependencies on each importer docs page should be elided into the usage with just a "gem install ..." right above the incantation. Right now it takes up a lot of the page.
1. I really missed having all the importers listed on the home page. Makes it easy for someone to know at a glance if their platform of choice is supported.

ashmaroli:
> this is in my queue to review..

No hurry, @parkr. This is still a WIP.
I just wanted your inputs on current *look-n-feel*.

Your *high-level thoughts* above is clear that you don't desire a redesign, but just want the CLI usage section to swap existing recommendation based on using Ruby API directly.

PR #507, opened 2022-12-16
Document command-line-interface of importers

ashmaroli:
A slimmer version of #503:
**Enhance just the dedicated page for importers.**


ashmaroli:
Deploy Preview URL:
https://deploy-preview-507--jekyll-import.netlify.app/

docs/_importers/behance.md
```diff
@@ -1,21 +1,7 @@
...
-
-To import your posts from your [Behance](http://behance.com), generate an API token for your user account and run:
-
-{% highlight bash %}
-$ ruby -r rubygems -e 'require "jekyll-import";
-    JekyllImport::Importers::Behance.run({
-      "user"      => "my_username",
-      "api_token" => "my_api_token"
-    })'
-{% endhighlight %}
```
parkr, line 19: The plugin you wrote is super cool and the final output is amazing! I'm kind of sad that a reader of this Jekyll site source wouldn't actually get to see much of the content that they'd view in the output. It simultaneously shows the power of Jekyll ("you can inject all sorts of content at build time!") and the drawback of that power ("a lot of the key content on the site isn't in the source files"). I wonder if it would make sense for either (1) us to generate the content into the site source which we push up to the repo or (2) the plugin to inject a variable into the Liquid template so our page can at least have a reference like `{{ auto_generated_usage_guide }}`. Some food for thought. Right now it's very magical but I'm debating if maybe it separates the source and final output too much?

docs/_importers/blogger.md
```diff
@@ -1,29 +1,13 @@
 ---
-layout: docs
-title: Blogger
 importer: true
 prev_section: behance
 link_source:  blogger
```
parkr, line 4: Side note: could we lexically alphabetize the list of importers and set these in a plugin or in Liquid too? This seems a bit inefficient to list them all here. When adding a new importer, the contributor has to know the order of their new importer in the context of other importers which we could easily get a computer to figure that out.

docs/_layouts/importer.html
```diff
@@ -0,0 +1,42 @@
...
+<h2>Invocation</h2>
+{% if page.prereq_info -%}
+<p>
+  {{ page.prereq_info | markdownify }}
+</p>
+{%- endif %}
+
+<p>
+  Sample snippet to invoke the importer:
+  <pre>jekyll import {{ page.cmd_name }}{% if page.cmd_opts %} {{ page.cmd_opts | map: 'switch' | join: ' ' }}{% endif %}</pre>
```
parkr, line 23: [S9Y Database importer has a load of params (19, to be exact). This makes this slightly harder to see since there's so much horizontal scrolling.](https://deploy-preview-507--jekyll-import.netlify.app/docs/s9ydatabase/) Is it at all possible to make this a multiline output?

```text
jekyll import s9ydatabase \
  --dbname DB \
  --socket SOCKET \
...
```

Or do you think that would take up too much space? 🤔

docs/_plugins/importer_metadata.rb
```diff
@@ -0,0 +1,73 @@
...
+        deps.sort!
+        deps
+      end
+    end
+
+    def options
+      @options ||= begin
+        @cmd.options.map do |o|
+          hsh = { "switch" => o.long }
+          if %r!(?<desc>.+?)(?:\z| \(default: (?<default_value>.*)\))! =~ o.description
```
parkr, line 52: Would be a good idea to add `o.default_value` or something to Mercenary, then we wouldn't need to rely on this. https://github.com/jekyll/mercenary/blob/master/lib/mercenary/option.rb

docs/_config.yml
```diff
@@ -12,6 +12,11 @@ collections:
     output: true
   importers:
     output: true
+    permalink: /docs/:path/
+
+defaults:
+  - { scope: { type: docs }, values: { layout: docs } }
+  - { scope: { type: importers }, values: { layout: importer } }
```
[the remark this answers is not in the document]
ashmaroli: Now that you mention it, the definition does look like *JSON*. But actually, it is just YAML.

But I shall change this to the suggested format for sure. :)

docs/_importers/behance.md
ashmaroli: For hosted metadata, we could generate another data file for CLI information using a Rake task just like we have one for the list of importer-dependencies.

But that is less exciting (for me, atleast) than having content generated automatically at runtime, especially because contributors / maintainers have to ensure the data file is updated as needed.
For example, say we add a new option to an importer or we make edits to the option-description, then
  - the data file has to be explicitly updated to ensure the change gets documented.
  - Maintainers would have to explicitly check if the pull request has the data file updated to match the proposed change before merging the PR.

Regarding the Liquid object `autogenerated...`, that too *is possible*, but I am not yet clear on the suggested use-case.

docs/_importers/blogger.md
ashmaroli: Technically, this is not needed at all. At least with modern Jekyll because basically there's already the concept of `next_doc` and `prev_doc` for `Jekyll::Document` objects.
These were removed in the other pull request that made wholesale changes to the docs but left it untouched here to maintain the narrow scope of current PR.

docs/_layouts/importer.html
ashmaroli: This block was just meant to quickly illustrate the CLI options. Technically for this importer, all of those options are *optional*. Therefore,
`jekyll import s9ydatabase` will work just as fine.
When making the block multiline, the end output is ugly (noisy) and we have to decide where exactly to draw the line to balance readability and accessibility. Besides, the table below lists all the options verbosely anyways.

docs/_plugins/importer_metadata.rb
ashmaroli: This is indeed a good idea. But is Mercenary still actively maintained?

docs/_importers/behance.md
parkr: Good points. Let's leave it how you have it now.

docs/_layouts/importer.html
parkr: Ok makes sense. I thought they were all required when reading the example invocation. If there's a default value for the flag, maybe we can use the common convention of wrapping the argument in square brackets like `--socket [SOCKET]`?

docs/_plugins/importer_metadata.rb
parkr: It's in the Jekyll org so we can modify it to include this! It hasn't gotten love recently since it did everything we wanted, but we could take some time to improve it. 😄

parkr (approved):

docs/_layouts/importer.html
ashmaroli: Modifying the long option here (i.e. bracketing the optional ARG) will cause a discrepancy between what's documented here and what gets generated in the terminal on invoking the CLI with `--help` switch.
Therefore, the best place to make this change is at the source-code level itself.

ashmaroli:
@jekyllbot: merge +doc

PR #508, opened 2022-12-19
Document importer dependencies as injected by local plugin

ashmaroli:
Use the importer-dependency metadata injected by `docs/_plugins/importer_metadata.rb` ( front matter key: `cmd_deps` ) to render importer dependency information at build-time instead of relying on manually generated data-file using a rake task.


ashmaroli:
@jekyllbot: merge +doc

PR #512, opened 2023-03-05
Re-implement Dotclear importer

ashmaroli:
Re-implement Dotclear importer based on export file provided by @jrfern in https://github.com/jekyll/jekyll-import/issues/510#issuecomment-1453747018.

This drops dependency on `activesupport`, includes associated tests and adds provided export file for future development.

Closes #510

PR #523, opened 2023-03-06
Refactor Dotclear importer methods into singleton instance methods

ashmaroli:
This change is purely for cosmetic reasons and have no functional changes.
The motivation behind this is to produce a saner diff in #512 which re-implements the importer, changing the Ruby API of the importer.


ashmaroli:
@jekyllbot: merge +dev

PR #527, opened 2023-05-07
add pebble importer

montdidier:
[Pebble](https://pebble.sourceforge.net) is a blog based with XML source files. Sadly it is EOL.

This script will enable folks to import their blog in to Jekyll, leveraging Nokogiri.

PR #534, opened 2024-01-17
Update docs: fix order of importers & add pebble importer

parkr:
Pebble importer was added in #527.


parkr:
@jekyllbot: merge +docs

PR #545, opened 2024-07-15
fix RuboCop

midzer:
Follow up to https://github.com/jekyll/jekyll-import/pull/544

PR #546, opened 2024-07-21
Fix erroneous cosmetic changes in Pebble importer

midzer:
(Trying) to fix https://github.com/jekyll/jekyll-import/pull/545

### DIFF
--- a/lib/jekyll-import/importers/pebble.rb
+++ b/lib/jekyll-import/importers/pebble.rb
@@ -28,11 +28,9 @@ module JekyllImport
       end
 
       def self.traverse_posts_within(directory, &block)
-        Dir.foreach(directory) do |fd|
+        Dir.each_child(directory) do |fd|
           path = File.join(directory, fd)
-          if [".", ".."].include?(fd)
-            next
-          elsif File.directory?(path)
+          if File.directory?(path)
             traverse_posts_within(path, &block)
           elsif path.end_with?("xml")
             yield(path) if block_given?
--- a/lib/jekyll-import/importers/pebble.rb
+++ b/lib/jekyll-import/importers/pebble.rb
@@ -15,7 +15,7 @@ module JekyllImport
       end
 
       def self.process(opts)
-        options = { directory => opts.fetch("directory", "") }
+        options = { :directory => opts.fetch("directory", "") }
 
         FileUtils.mkdir_p("_posts")
         FileUtils.mkdir_p("_drafts")
@@ -30,7 +30,7 @@ module JekyllImport
       def self.traverse_posts_within(directory, &block)
         Dir.foreach(directory) do |fd|
           path = File.join(directory, fd)
-          if fd.include?(".") || fd.include?("..")
+          if [".", ".."].include?(fd)
             next
           elsif File.directory?(path)
             traverse_posts_within(path, &block)


lib/jekyll-import/importers/pebble.rb
```diff
@@ -15,7 +15,7 @@ def self.specify_options(c)
       end
 
       def self.process(opts)
-        options = { directory => opts.fetch("directory", "") }
+        options = { :directory => opts.fetch("directory", "") }
```
parkr, line 18: Great work here! I saw that too (when I looked again). The key needs to be a symbol (used a colon `:`).

lib/jekyll-import/importers/pebble.rb
```diff
@@ -30,7 +30,7 @@ def self.process(opts)
       def self.traverse_posts_within(directory, &block)
         Dir.foreach(directory) do |fd|
           path = File.join(directory, fd)
-          if fd.include?(".") || fd.include?("..")
+          if [".", ".."].include?(fd)
```
parkr, line 33: Take a closer look at the old code:

```ruby
if fd == '.' || fd == '..'
```

parkr (commented):
One looks right to me but the other needs another look. Thanks for playing along!

lib/jekyll-import/importers/pebble.rb
midzer: Thanks for your reply.

Yes, I'm new to Ruby. I tried to understand

```
# bad
a = 'a'
foo if a == 'a' || a == 'b' || a == 'c'

# good
a = 'a'
foo if ['a', 'b', 'c'].include?(a)
```

via https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MultipleComparison

But I don't get it...

parkr: No worries at all! Consider that when you iterate through the entries in a directory you get all the file descriptors accessible in that directory. In Unix, that includes the current directory `.` and parent directory `..`. You can see them when you run the command `ls -a`. If you have the following entries in a directory:

```
.
..
MyScan.pdf
Foo
```

Each name will be run through your code. The goal is to skip the current directory `.` and parent directory `..` **only**. Additionally, all object allocations in a loop like this (for example, creating a new array) are costly, so we like to avoid them.

Try running your code with the entries above. You want to skip only the current & parent directories, but keep `MyScan.pdf` and `Foo`. 

With Ruby installed on your computer, you can use the `irb` command to start an interactive Ruby REPL. This will allow you to run your code and modify it quickly.

```ruby
input_entries = [".", "..", "MyScan.pdf", "Foo"]
input_entries.each do |fd|
  puts "matched #{fd}" if YOUR CODE
end
```

midzer: Thanks for your detailed explanation, @parkr 

I ran via `irb`

```
2.5.3 :001 > input_entries = [".", "..", "MyScan.pdf", "Foo"]
 => [".", "..", "MyScan.pdf", "Foo"] 
2.5.3 :002 > input_entries.each do |fd|
2.5.3 :003 >     if [".", ".."].include?(fd)
2.5.3 :004?>         next
2.5.3 :005?>     end
2.5.3 :006?>     puts "matched #{fd}"
2.5.3 :007?> end
matched MyScan.pdf
matched Foo
 => [".", "..", "MyScan.pdf", "Foo"]
```

and

```
2.5.3 :001 > input_entries = [".", "..", "MyScan.pdf", "Foo"]
 => [".", "..", "MyScan.pdf", "Foo"] 
2.5.3 :002 > input_entries.each do |fd|
2.5.3 :003 >     next if [".", ".."].include?(fd)
2.5.3 :004?>     puts "matched #{fd}"
2.5.3 :005?> end
matched MyScan.pdf
matched Foo
 => [".", "..", "MyScan.pdf", "Foo"]
```
 and both ways appear to match desired file names.

I digged a little deeper in the issue and found out, that https://ruby-doc.org/3.2.2/Dir.html#method-c-each_child is probably more suitable for the loop than a manual check:

> Calls the block once for each entry except for “.” and “..” in the named directory, passing the filename of each entry as a parameter to the block.

ashmaroli:
> and both ways appear to match desired file names.

@midzer :) the thing is that both forms are 99% equivalent. The gotcha with Ruby compared to Python is that one can have multiple ways to implement the same idea. Therefore, `if something; next item; end` is almost equivalent to `next item if something`.

What Parker highlighted in his explanation is that creating the array `[".", ".."]` for each iteration is waste of resources, albeit hardly noticeable in most situations. To that end, your suggestion to use `Dir.each_child` is indeed a superior alternative. It's just that we would have to be certain that subdirectories are handled properly as well (_Disclaimer: I'm not totally familiar with the pebble platform or the importer code_).

lib/jekyll-import/importers/pebble.rb
parkr: `Dir.each_child` would probably be a better option, yes! It looks like it is available as of Ruby 2.5, so we can use it: https://www.bigbinary.com/blog/ruby-2_5-introduces-dir-children-and-dir-each_child

parkr (approved):

ashmaroli:
LGTM!

@jekyllbot: merge +dev

ashmaroli:
Thank you @midzer. :)

PR #551, opened 2024-10-09
550 drupal 7 does not import author name email

### LINKED ISSUE #550: Drupal 7 does not include author name/email in posts
OS: M1 apple chip macos, Sonoma 14.7
Ruby version: 3.1.3p185 (2022-11-24 revision 1a6b16756e) [arm64-darwin23]
Jekyll version: 3.10.0
Jekyll import version: 0.24.0

Importer: drupal7

I am importing posts (nodes) from a Drupal 7 database.  The local version of this DB is running in a Docker container, MySql 8.0.39.

Because the Drupal 7 blog contained thousands of posts from hundreds of authors, I need the author information.  This can be found in Drupal's `users` table, and is cross-referenced by the `uid` of the node.  I am not a Drupal expert but I believe this relationship (between the node and users table) is reliable in all cases, and that the users table should contain at least a name and email.

For my purpose, I only need `u.name` (where u is an alias to the users table, and the name value is in the format `first.last`).  I can see how `u.mail` (which is email address) would also be helpful, though.

I know how to fix this, and I will make a quick PR.  However, I am also not a ruby developer, and the instructions to test don't work for me because
1. I cannot find `script/console` as described in the [Contributing page](https://import.jekyllrb.com/docs/contributing/) and
2. I get a `make` error for dependency `hpricot-0.8.6`.  This is likely due to the fact that I'm on an M1 mac, or I need to do a custom install/build of hpricot, or something.

Given the simple nature of the change and the testing issues I'm facing, I am hoping that someone can guide me through the process of testing this, or provide a build for me to install and test. 🙏

lyndseypadget:
Fixes #550.

While the issue was written up for Drupal 7, I realized that the drupal7 and drupal8 importers share a common file.  I believe this approach should work for Drupal 8 as well, and it's reasonable to assume that users would want author name and email supported in both versions.

As mentioned in the issue description, I could use some guidance on how to test this.  Or, provide a build and I'll be the guinea pig!  (For Drupal 7 at least.)

### DIFF
--- a/lib/jekyll-import/importers/drupal7.rb
+++ b/lib/jekyll-import/importers/drupal7.rb
@@ -34,10 +34,14 @@ SQL
                        n.created,
                        n.status,
                        n.type,
+                       u.name,
+                       u.mail,
                        #{tag_group}
                 FROM #{prefix}node AS n
                 LEFT JOIN #{prefix}field_data_body AS fdb
                   ON fdb.entity_id = n.nid AND fdb.entity_type = 'node'
+                RIGHT JOIN #{prefix}users AS u
+                  ON u.uid = n.uid
                 WHERE (#{types})
 QUERY
 
--- a/lib/jekyll-import/importers/drupal8.rb
+++ b/lib/jekyll-import/importers/drupal8.rb
@@ -34,10 +34,14 @@ SQL
                        n.created,
                        n.status,
                        n.type,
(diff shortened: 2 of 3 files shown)


parkr:
LGTM, and thank you for testing!

@jekyllbot: merge +minor

PR #568, opened 2026-04-02
Add SLSA generic generator workflow

pathum2583-eng:
This workflow generates SLSA provenance files for projects, satisfying level 3 requirements. It includes steps for building artifacts and generating subject hashes.


mattr-:
Ruby projects that publish gems have other ways of doing SLSA provenance via GitHub Actions and Rubygems so I don't think change would apply here. Closing.

PR #573, opened 2026-06-21
rss: fix crash if missing item.enclosure & add tests

parkr:
Fixes a bug where an `<enclosure>` was expected for every `<item>`, which could break importing of certain sites.

Adds test/test_rss_importer.rb covering validation, post generation, tagging (explicit --tag and --extract_tags), canonical links, and audio rendering, plus a clean test/mocks/rss_feed.xml fixture.

✨ Claude Opus 4.8 was used in the writing of this test.


parkr:
@jekyllbot: merge +minor

